PR #38, opened 2020-11-28
Progress bar

xwu64:
I've updated the implementation of progress bar. The implementation is replaced by library [tqdm](https://github.com/tqdm/tqdm). 

After using `tqdm`, the progress bar is always enabled. Could you try to run some simulations? I'd like to hear your feedback about this update.


Alex-Kolar:
I think it looks good. I just made some minor changes to the timeline file and fixed the length of the progress bar at 50 characters.

Alex-Kolar (approved):

PR #45, opened 2020-12-12
Rework client close/kill methods; fix pickling issue

Alex-Kolar:


xwu64:
I noticed three new files in the `utils` folder. They seem like the test code. Should we exclude them from the repository?

PR #51, opened 2020-12-15
Merge master into parall branch

xwu64:


Alex-Kolar:
In reviewing the merge conflicts, I think some changes may be committed to the master branch that would break the code, such as adding non-existent files to __init__ and changing the import type to some functions in the test files (namely set_ends using node name strings instead of nodes). I created a master_clone branch and I suggest we merge from here instead of master.

xwu64:
> In reviewing the merge conflicts, I think some changes may be committed to the master branch that would break the code, such as adding non-existent files to **init** and changing the import type to some functions in the test files (namely set_ends using node name strings instead of nodes). I created a master_clone branch and I suggest we merge from here instead of master.

Event we merge master_clone branch into parallel branch, we also need to address these conflicts, right?
For conflicting files, could we adopt code in the parallel branch? Because we are merging the master branch into parallel branch, some code in the master branch may be out of date.

Alex-Kolar:
Yes, we will still need to address the conflicts. However, the conflict resolutions will be committed to the master_clone branch rather than the master branch. Some changes from the parallel branch will break the test files. Namely, the new QuantumChannel.add_ends method can take a string as an argument, while this will not work with the channel code on master.

xwu64:
Because we are merging the master branch in the parallel branch, all changes in the merging will be committed to the parallel branch. The master branch should not be affected by this merge operation, right?

Alex-Kolar:
When the merge conflicts are resolved, they are commited to the source branch before merging.

xwu64:
Ok. If it's caused by the Github? Could we merge two branches locally and push to the repo? Creating a new branch may be also an acceptable solution. You can choose a way you want.

Alex-Kolar:
I think it's a feature of git. I have made the new branch, and I will open a new pull request.

PR #63, opened 2021-01-26
Added random coherence time support in Memory class

viachesl:
Now it's possible to pass both average time (as before) or tuple with the average time and its standard deviation. I don't know how to organize correctly namespace in test_memory.py but I hope its current version is not too far from what you need.

### DIFF
--- a/src/components/memory.py
+++ b/src/components/memory.py
@@ -9,6 +9,7 @@ from math import sqrt, inf
 from typing import Any, List, TYPE_CHECKING, Dict
 
 from numpy import random
+from scipy import stats
 
 if TYPE_CHECKING:
     from ..entanglement_management.entanglement_protocol import EntanglementProtocol
@@ -107,6 +108,7 @@ class Memory(Entity):
         frequency (float): maximum frequency at which memory can be excited.
         efficiency (float): probability of emitting a photon when excited.
         coherence_time (float): average usable lifetime of memory (in seconds).
+                                or ( time_average(float), time_stdev(float) ) tuple
         wavelength (float): wavelength (in nm) of emitted photons.
         qstate (QuantumState): quantum state of memory.
         entangled_memory (Dict[str, Any]): tracks entanglement state of memory.
@@ -116,7 +118,7 @@ class Memory(Entity):
     _meas_circuit.measure(0)
 
     def __init__(self, name: str, timeline: "Timeline", fidelity: float, frequency: float,
-                 efficiency: float, coherence_time: int, wavelength: int):
+                 efficiency: float, coherence_time: Any, wavelength: int):
         """Constructor for the Memory class.
 
         Args:
@@ -125,7 +127,8 @@ class Memory(Entity):
             fidelity (float): fidelity of memory.
             frequency (float): maximum frequency of excitation for memory.
             efficiency (float): efficiency of memories.
-            coherence_time (float): average time (in s) that memory state is valid.
+            coherence_time (float): average time (in s) that memory state is valid
+                                    or ( time_average(float), time_stdev(float) ) tuple
             wavelength (int): wavelength (in nm) of photons emitted by memories.
             qstate_key (int): key for associated quantum state in timeline's quantum manager.
         """
@@ -138,7 +141,23 @@ class Memory(Entity):
         self.raw_fidelity = fidelity
         self.frequency = frequency
         self.efficiency = efficiency
-        self.coherence_time = coherence_time  # coherence time in seconds
+        # coherence time in seconds
+        if not type( coherence_time ) is tuple:
+            self.coherence_time = coherence_time
+            self.random_coherence_time = False
+        else:
+            self.coherence_time = coherence_time[0]
+            self.coherence_time_stdev = coherence_time[1]
+            self.random_coherence_time = ( self.coherence_time_stdev > 0.0 and
+                                          self.coherence_time > 0.0 )
+            
+        if self.random_coherence_time:
+            self.coherence_time_distribution = lambda: stats.truncnorm.rvs(
+                -0.95 * self.coherence_time / self.coherence_time_stdev,
+                19.0 * self.coherence_time / self.coherence_time_stdev,
+                self.coherence_time,
+                self.coherence_time_stdev )
+        
         self.wavelength = wavelength
         self.qstate_key = timeline.quantum_manager.new()
 
@@ -257,8 +276,13 @@ class Memory(Entity):
     def _schedule_expiration(self) -> None:
         if self.expiration_event is not None:
             self.timeline.remove_event(self.expiration_event)
+            
+        if not self.random_coherence_time:
+            coherence_period = self.coherence_time
+        else:
+            coherence_period = self.coherence_time_distribution()
 
-        decay_time = self.timeline.now() + int(self.coherence_time * 1e12)
+        decay_time = self.timeline.now() + int(coherence_period * 1e12)
         process = Process(self, "expire", [])
         event = Event(decay_time, process)
         self.timeline.schedule(event)
--- a/src/components/memory.py
+++ b/src/components/memory.py
@@ -297,8 +297,12 @@ class Memory(Entity):
 class MemoryWithRandomCoherenceTime(Memory):
     """Individual single-atom memory.
 
-    This class models a single-atom memory, where the quantum state is stored as the spin of a single ion.
-    This class will replace the older implementation once completed.
+    This class inherits Memory class and provides possibility to use stochastic model of
+    coherence time. This means that loss of entanglement of the memory with a photon occurs
+    at random time given by truncated normal distribution with average value set by
+    'coherence_time' input parameter and with standard deviation set by 'coherence_time_stdev'
+    input parameter. If coherence_time_stdev <= 0.0 is passed, the class behaves exactly as
+    original Memory class.
 
     Attributes:
         name (str): label for memory instance.
--- a/src/components/memory.py
+++ b/src/components/memory.py
@@ -94,7 +94,6 @@ class MemoryArray(Entity):
     def set_node(self, node: "QuantumRouter") -> None:
         self.owner = node
 
-
 class Memory(Entity):
     """Individual single-atom memory.
 
@@ -108,7 +107,6 @@ class Memory(Entity):
         frequency (float): maximum frequency at which memory can be excited.
         efficiency (float): probability of emitting a photon when excited.
         coherence_time (float): average usable lifetime of memory (in seconds).
-                                or ( time_average(float), time_stdev(float) ) tuple
         wavelength (float): wavelength (in nm) of emitted photons.
         qstate (QuantumState): quantum state of memory.
         entangled_memory (Dict[str, Any]): tracks entanglement state of memory.
@@ -118,7 +116,7 @@ class Memory(Entity):
     _meas_circuit.measure(0)
 
     def __init__(self, name: str, timeline: "Timeline", fidelity: float, frequency: float,
-                 efficiency: float, coherence_time: Any, wavelength: int):
+                 efficiency: float, coherence_time: int, wavelength: int):
         """Constructor for the Memory class.
 
         Args:
@@ -127,8 +125,7 @@ class Memory(Entity):
             fidelity (float): fidelity of memory.
             frequency (float): maximum frequency of excitation for memory.
             efficiency (float): efficiency of memories.
-            coherence_time (float): average time (in s) that memory state is valid
-                                    or ( time_average(float), time_stdev(float) ) tuple
+            coherence_time (float): average time (in s) that memory state is valid.
             wavelength (int): wavelength (in nm) of photons emitted by memories.
             qstate_key (int): key for associated quantum state in timeline's quantum manager.
         """
@@ -141,23 +138,7 @@ class Memory(Entity):
         self.raw_fidelity = fidelity
         self.frequency = frequency
         self.efficiency = efficiency
-        # coherence time in seconds
-        if not type( coherence_time ) is tuple:
-            self.coherence_time = coherence_time
-            self.random_coherence_time = False
-        else:
-            self.coherence_time = coherence_time[0]
-            self.coherence_time_stdev = coherence_time[1]
-            self.random_coherence_time = ( self.coherence_time_stdev > 0.0 and
-                                          self.coherence_time > 0.0 )
-            
-        if self.random_coherence_time:
-            self.coherence_time_distribution = lambda: stats.truncnorm.rvs(
-                -0.95 * self.coherence_time / self.coherence_time_stdev,
-                19.0 * self.coherence_time / self.coherence_time_stdev,
-                self.coherence_time,
-                self.coherence_time_stdev )
-        
+        self.coherence_time = coherence_time  # coherence time in seconds
         self.wavelength = wavelength
         self.qstate_key = timeline.quantum_manager.new()
 
@@ -276,13 +257,8 @@ class Memory(Entity):
     def _schedule_expiration(self) -> None:
         if self.expiration_event is not None:
             self.timeline.remove_event(self.expiration_event)
-            
-        if not self.random_coherence_time:
-            coherence_period = self.coherence_time
-        else:
-            coherence_period = self.coherence_time_distribution()
 
-        decay_time = self.timeline.now() + int(coherence_period * 1e12)
+        decay_time = self.timeline.now() + int(self.coherence_time * 1e12)
         process = Process(self, "expire", [])
         event = Event(decay_time, process)
         self.timeline.schedule(event)
@@ -317,4 +293,70 @@ class Memory(Entity):
     def detach(self, observer: 'EntanglementProtocol'):
         if observer in self._observers:
             self._observers.remove(observer)
+
+class MemoryWithRandomCoherenceTime(Memory):
+    """Individual single-atom memory.
+
+    This class models a single-atom memory, where the quantum state is stored as the spin of a single ion.
+    This class will replace the older implementation once completed.
+
+    Attributes:
+        name (str): label for memory instance.
+        timeline (Timeline): timeline for simulation.
+        fidelity (float): (current) fidelity of memory.
+        frequency (float): maximum frequency at which memory can be excited.
+        efficiency (float): probability of emitting a photon when excited.
+        coherence_time (float): average usable lifetime of memory (in seconds).
+        coherence_time_stdev (float): standard deviation of coherence time
+        wavelength (float): wavelength (in nm) of emitted photons.
+        qstate (QuantumState): quantum state of memory.
+        entangled_memory (Dict[str, Any]): tracks entanglement state of memory.
+    """
+
+    def __init__(self, name: str, timeline: "Timeline", fidelity: float, frequency: float,
+                 efficiency: float, coherence_time: float, coherence_time_stdev: float, wavelength: int):
+        """Constructor for the Memory class.
+
+        Args:
+            name (str): name of the memory instance.
+            timeline (Timeline): simulation timeline.
+            fidelity (float): fidelity of memory.
+            frequency (float): maximum frequency of excitation for memory.
+            efficiency (float): efficiency of memories.
+            coherence_time (float): average time (in s) that memory state is valid
+            coherence_time_stdev (float): standard deviation of coherence time
+            wavelength (int): wavelength (in nm) of photons emitted by memories.
+            qstate_key (int): key for associated quantum state in timeline's quantum manager.
+        """
+
+        super(MemoryWithRandomCoherenceTime, self).__init__(name, timeline, fidelity, frequency, 
+                         efficiency, coherence_time, wavelength)
+        
+        # coherence time standard deviation in seconds
+        self.coherence_time_stdev = coherence_time_stdev
+        self.random_coherence_time = ( coherence_time_stdev > 0.0 and
+                                      self.coherence_time > 0.0 )
+        
+    def coherence_time_distribution(self) -> None:
+        return stats.truncnorm.rvs(
+            -0.95 * self.coherence_time / self.coherence_time_stdev,
+            19.0 * self.coherence_time / self.coherence_time_stdev,
+            self.coherence_time,
+            self.coherence_time_stdev )        
+
+    def _schedule_expiration(self) -> None:
+        if self.expiration_event is not None:
+            self.timeline.remove_event(self.expiration_event)
+            
+        coherence_period = ( self.coherence_time_distribution() 
+                            if self.random_coherence_time else 
+                            self.coherence_time )
+
+        decay_time = self.timeline.now() + int(coherence_period * 1e12)
+        process = Process(self, "expire", [])
+        event = Event(decay_time, process)
+        self.timeline.schedule(event)
+
+        self.expiration_event = event
+
 
--- a/src/components/memory.py
+++ b/src/components/memory.py
@@ -94,7 +94,6 @@ class MemoryArray(Entity):
     def set_node(self, node: "QuantumRouter") -> None:
         self.owner = node
 
-
 class Memory(Entity):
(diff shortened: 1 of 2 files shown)


xwu64:
We're appreciate your contribution about the class `Memory`. I've two ideas about this change. And I'd like to hear feedback from you. 

---

1. Could we create a new class that inherits class `Memory`, like `MemoryWithRandCoherenceTime`, to support this new feature? 

There are three reasons about creating a new class:

1. A separate class could reduce the complexity of class `Memory`.
2. People who is using class `Memory` will not be affected by this change.
3. We could avoid conflict when we merge some commits in the future.

----
2. Could we avoid creating a new class and implementing the random coherence time in a better way? 

I noticed two arguments (coherence time and coherence time standard deviation) are required for generating the distribution of coherence time. If we set the standard deviation to 0, could we get a fixed coherence time? If it's true, maybe we could make the default value of this argument as 0. Then, people could choose whether to use the random coherence time by setting the standard deviation. 

---

I also have a silly question about the formula that calculating the coherence time distribution. If the number `-0.95` and `19.0` are the suitable coefficients for any memory? 

**Thanks for your contribution again!**

viachesl:
Hi xwu64,

OK, I will transform this changes into a new class inheriting Memory, I got it why it's needed now. Now answer on your question:

> I also have a silly question about the formula that calculating the coherence time distribution. If the number -0.95 and 19.0 are the suitable coefficients for any memory?

This function
```
        if self.random_coherence_time:
            self.coherence_time_distribution = lambda: stats.truncnorm.rvs(
                -0.95 * self.coherence_time / self.coherence_time_stdev,
                19.0 * self.coherence_time / self.coherence_time_stdev,
                self.coherence_time,
                self.coherence_time_stdev )
```
Setups normal distribution with parameters self.coherence_time and self.coherence_time_stdev as usual, but truncated to avoid negative and zero values. I don't know how to set up infinite upper bound in that function, so I set it up 20 * self.coherence_time. Lower bound is set up in this code as 0.05 * self.coherence_time. I think it's reasonable. But what do you think on that?

P.S. Maybe it's better to set up the coefficients as -0.99/99.0, but I don't know how it will affect the performance. Now with -0.95/19.0 upgrated Memory class is about 25% slower than old one.

xwu64:
1. I'm not sure a reasonable range of coherence time. Do you have any reference about the range of coherence time between 0.05*self.coherence_time and 20*self.coherence_time? Is it reasonable to set them as the argument of constructor function? 
2. Is this distribution suitable for any type of memory? If this distribution only suitable a part of quantum memory, maybe you could name the class by the type of memory. 
3. The potential problem of using tuple as argument is the usability of code. For user who does not read the code, they could not know the syntax used for this argument. We may want to expose these information.  
4. To avoid a descendant class, we could set the default value of `self.coherence_time_stdev` to `0`. The randomness of coherence time will only happen when user sets the argument to a value > 0. 

I prefer to add an argument `coherence_time_stdev` in the constructor function and set the default value as `0` if the coherence time of all memory follows such distribution.

src/components/memory.py
```diff
@@ -138,7 +141,23 @@ def __init__(self, name: str, timeline: "Timeline", fidelity: float, frequency:
...
-        self.coherence_time = coherence_time  # coherence time in seconds
+        # coherence time in seconds
+        if not type( coherence_time ) is tuple:
+            self.coherence_time = coherence_time
+            self.random_coherence_time = False
+        else:
+            self.coherence_time = coherence_time[0]
+            self.coherence_time_stdev = coherence_time[1]
+            self.random_coherence_time = ( self.coherence_time_stdev > 0.0 and
+                                          self.coherence_time > 0.0 )
```
LesterKim:
```suggestion
        if type(coherence_time) is tuple:
            self.coherence_time, self.coherence_time_stdev = coherence_time
            self.random_coherence_time = self.coherence_time_stdev > 0.0 < self.coherence_time
        else:
            self.coherence_time = coherence_time
            self.random_coherence_time = False
```

src/components/memory.py
```diff
@@ -138,7 +141,23 @@ def __init__(self, name: str, timeline: "Timeline", fidelity: float, frequency:
...
+            
+        if self.random_coherence_time:
+            self.coherence_time_distribution = lambda: stats.truncnorm.rvs(
+                -0.95 * self.coherence_time / self.coherence_time_stdev,
+                19.0 * self.coherence_time / self.coherence_time_stdev,
+                self.coherence_time,
+                self.coherence_time_stdev )
```
LesterKim: This should be a separate method.

src/components/memory.py
```diff
@@ -257,8 +276,13 @@ def update_state(self, state: List[complex]) -> None:
     def _schedule_expiration(self) -> None:
         if self.expiration_event is not None:
             self.timeline.remove_event(self.expiration_event)
+            
+        if not self.random_coherence_time:
+            coherence_period = self.coherence_time
+        else:
+            coherence_period = self.coherence_time_distribution()
```
LesterKim:
```suggestion
        coherence_period = self.coherence_time_distribution() if self.random_coherence_time else self.coherence_time
```

tests/components/test_memory.py
```diff
@@ -185,3 +185,50 @@ def test_Memory__schedule_expiration():
...
+        tl.init()
+        tl.run()
+        assert times_of_expiration_calculated[i] == tl.now()
+        
+    sumX = times_of_expiration_calculated[0]
+    sumXX = times_of_expiration_calculated[0]**2
+    for i in range( 1, len( times_of_expiration_calculated ) ):
+        x = times_of_expiration_calculated[i]-times_of_expiration_calculated[i-1]
+        sumX += x
+        sumXX += x*x
```
LesterKim: What is a more descriptive name for `x` and `X`?

tests/components/test_memory.py
```diff
@@ -185,3 +185,50 @@ def test_Memory__schedule_expiration():
...
+    
+    times_of_expiration_calculated = [0]
+    np.random.seed(2)
+    for i in range( NUM_TRIALS ):
+        times_of_expiration_calculated.append( times_of_expiration_calculated[-1]
+                                              + int(mem.coherence_time_distribution()*1e12) )
+    times_of_expiration_calculated.pop(0)
+    
+    np.random.seed(2)
+    process = Process(mem, "update_state", [[complex(math.sqrt(1/2)), complex(math.sqrt(1/2))]])
```
LesterKim, line 207:
```suggestion
    process = Process(mem, "update_state", [2 * [complex(math.sqrt(1/2))]])
```

tests/components/test_memory.py
```diff
@@ -185,3 +185,50 @@ def test_Memory__schedule_expiration():
...
+        sumX += x
+        sumXX += x*x
+    
+    avg_simulated = sumX / NUM_TRIALS * 1e-12
+    stdev_simulated = np.sqrt( ( sumXX - sumX * sumX * 1.0/NUM_TRIALS ) / NUM_TRIALS )*1e-12
+
+    #print( 'input avg. =', coherence_period_avg )
+    #print( 'input st. dev. =', coherence_period_stdev )
+    #print( 'simulated avg. =', avg_simulated )
+    #print( 'simulated st. dev. =', stdev_simulated )
```
LesterKim: Why the comments?
```suggestion
```

LesterKim (changes requested):

tests/components/test_memory.py
viachesl:
```
x -> period
sumX -> sumPeriod
sumXX -> sumPeriodSqr
```
Do you mean this?

src/components/memory.py
viachesl: How can I update my push request? Just edit files on my PC and make one more commit? Can I update my files with the changes You have done and I confirmed?

src/components/memory.py
viachesl: Your colleague suggested to use extra variable like coherence_time_stdev = 0 (default value is zero that means fixed time, without random values). How can I change this after I confirm this your edits?

tests/components/test_memory.py
LesterKim: Ok. Then, please update `x` to `period`, `sumX` to `period_sum`, and `sumXX` to `period_squared_sum`.

src/components/memory.py
LesterKim:
```python
            self.coherence_time, self.coherence_time_stdev = coherence_time
```
is same as
```python
            self.coherence_time = coherence_time[0]
            self.coherence_time_stdev = coherence_time[1]
```

src/components/memory.py
LesterKim: Yes.

viachesl:
Sorry for so long delay, I was busy. I think I satisfied all the suggestions said in this topic. Here is the code using this new memory class:
```
import sys, time
sys.path.append( '../' )

from sequence.topology.node import Node
from sequence.components.memory import MemoryWithRandomCoherenceTime
from sequence.entanglement_management.generation import EntanglementGenerationA
from sequence.kernel.timeline import Timeline
from sequence.topology.node import BSMNode
from sequence.components.optical_channel import QuantumChannel, ClassicalChannel
from sequence.entanglement_management.entanglement_protocol import EntanglementProtocol

import numpy as np

doSimulations = True

class SimpleManager():
    def __init__(self):
        self.raw_counter = 0
        self.ent_counter = 0

    def update(self, protocol, memory, state):
        if state == 'RAW':
            self.raw_counter += 1
            memory.reset()
        else:
            self.ent_counter += 1
            memory.reset()

class EntangleGenNode(Node):
    def __init__(self, name: str, tl: Timeline, **memory_params ):
        super().__init__(name, tl)
        #fidelity=0.9, frequency=2000, efficiency=1, coherence_time=-1, wavelength=500 
        self.memory = MemoryWithRandomCoherenceTime('%s.memo'%name, tl, **memory_params ) #0.9, 2000, 1, -1, 500)
        self.memory.owner = self
        self.resource_manager = SimpleManager()
        self.protocols = []

    def receive_message(self, src: str, msg: "Message") -> None:
        self.protocols[0].received_message(src, msg)

    def create_protocol(self, middle: str, other: str):
        self.protocols = [EntanglementGenerationA(self, '%s.eg'%self.name, middle, other, self.memory)]

def pair_protocol( p1: EntanglementProtocol, p2: EntanglementProtocol ):
    p1.set_others( p2 )
    p2.set_others( p1 )

def runSimulations( _distance, _memory_params, _detector_params, _qchannel_params, _stochastic, _repetitions ):
    __seed = int( time.time() * 1000 ) % 2**32
    np.random.seed( __seed )
    
    tl = Timeline()
    tl.seed( __seed % 256 )
    #light_velocity = 2e8 #m/s
    
    node1 = EntangleGenNode( 'node1', tl, **_memory_params )
    node2 = EntangleGenNode( 'node2', tl, **_memory_params )
    bsm_node = BSMNode('bsm_node', tl, ['node1', 'node2'])
    
    for name, param in _detector_params.items():
        bsm_node.bsm.update_detectors_params( name, param )
        #bob.update_detector_params(i, name, param)
    #bsm_node.bsm.update_detectors_params( 'efficiency', 1 )
    
    _qchannel_params[ "distance" ] = _distance/2
    
    qc1 = QuantumChannel( 'qc1', tl, **_qchannel_params )
    qc2 = QuantumChannel( 'qc2', tl, **_qchannel_params )
    qc1.set_ends(node1, bsm_node)
    qc2.set_ends(node2, bsm_node)
    
    nodes = [ node1, node2, bsm_node ]
    
    for i in range(3):
        for j in range(3):
            if i == j:
                continue
            cc = ClassicalChannel( 'cc_%s_%s' % (nodes[i].name, nodes[j].name),
                                   tl, abs(i-j)*_distance/(len(nodes)-1), -1 )
            cc.set_ends( nodes[i], nodes[j] )
    
    #print( 'tl.now() = %.3f ms' % ( tl.now() / 1e9 ) )
    for i in range( _repetitions ):
        #tl.time = tl.now() + 1e11
        node1.create_protocol('bsm_node', 'node2')
        node2.create_protocol('bsm_node', 'node1')
        pair_protocol(node1.protocols[0], node2.protocols[0])

        node1.memory.reset()
        node2.memory.reset()

        tl.init()

        node1.protocols[0].start()
        node2.protocols[0].start()

        tl.run()
        #print( 'tl.now() = %.3f ms' % ( tl.now() / 1e9 ) )
    #print( 'tl.now() = %.3f ms' % ( tl.now() / 1e9 ) )

    #print(node1.resource_manager.ent_counter, ':', node1.resource_manager.raw_counter)
    # (around 500:500; the exact number depends on the seed of numpy.random)
    #fidelities = np.array( node1.resource_manager.fidelities )
    #numPts = fidelities.shape[0]
    #xSum = np.sum( fidelities )
    #xxSum = np.sum( fidelities**2 )
    #xAvg = xSum / numPts
    #xStDev = np.sqrt( abs( xxSum / numPts - xAvg**2 ) )
    #print( 'Average fidelity: %f\n fidelity st dev: %f' % ( xAvg, xStDev ) )
    return node1.resource_manager.ent_counter / tl.now() * 1e12

if doSimulations:
    _t0 = time.time()
    N = 1000
    stochastic = True
    Ls = np.arange( 1, 101, 2 ) * 1e3
    coherence_time_avg = 0.1e-3 #0.55 * Ls[ len( Ls ) // 2 ] / 2e8
    coherence_time_stdev = 0.2 * coherence_time_avg if stochastic else 0.0
    attenuation = 0.0002
    mem_eff = 0.9
    det_eff = 0.8
    
    if False:
        print( 'coherence time = %.3f ms' % ( coherence_time_avg / 1e-3 ) )
        memory_params = { "fidelity": 1.0, "frequency": 2e6, "efficiency": mem_eff,
                          "coherence_time": coherence_time_avg,
                          "coherence_time_stdev": coherence_time_stdev, "wavelength": 500 }
        detector_params = { "efficiency": det_eff, 
                           #"dark_count": 10, 
                           "time_resolution": 150, 
                           "count_rate": 50e7 
                           }
        qchannel_params = { "attenuation": attenuation, 
                           "polarization_fidelity": 1.0, 
                           "distance": 0 }
        rs = np.zeros( Ls.shape )
        for i in range( len( Ls ) ):
            rs[i] = runSimulations( Ls[i], memory_params, detector_params, qchannel_params, stochastic, N )
        #print( 'entanglements per second generated: %.3f' % etngs_per_sec )
        #print( rs )
        fname = ( 'entRate(%satt=%.3fdBkm,cohTime=%s%smemEff=%.2f,detEff=%.2f,N=1E%d).txt' 
                 % ( 'stoch,' if stochastic and coherence_time_avg > 0 and coherence_time_stdev > 0 else '',
                     attenuation * 1000,
                     '%.3fms' % ( coherence_time_avg / 1e-3 ) if coherence_time_avg > 0.0 else 'inf',
                     ( '+-%.2fmcs,' % ( coherence_time_stdev / 1e-6 )
                       if ( stochastic and coherence_time_avg > 0 and coherence_time_stdev > 0 )
                       else ',' ),
                     mem_eff, det_eff, np.log10(N) ) )
        np.savetxt( fname, np.concatenate( ( Ls.reshape( (Ls.shape[0],1) ),
                                            rs.reshape( (Ls.shape[0],1) ) ), axis = 1 ) )
        print( 'Calculated in %.3f seconds' % ( time.time() - _t0 ) )
    if True:
        import matplotlib.style
        matplotlib.style.use('classic')
        import matplotlib.pyplot as plt
        from matplotlib.ticker import MultipleLocator

        stochastic = True
        attenuation = 0.0
        coherence_time_avg = -1
        fname1 = ( 'entRate(%satt=%.3fdBkm,cohTime=%s%smemEff=%.2f,detEff=%.2f,N=1E%d).txt' 
                 % ( 'stoch,' if stochastic and coherence_time_avg > 0.0 and coherence_time_stdev > 0
                     else '',
                     attenuation * 1000,
                     '%.3fms' % ( coherence_time_avg / 1e-3 ) if coherence_time_avg > 0.0 else 'inf',
                     ( '+-%.2fmcs,' % ( coherence_time_stdev / 1e-6 )
                       if ( stochastic and coherence_time_avg > 0 and coherence_time_stdev > 0 )
                       else ',' ),
                     mem_eff, det_eff, np.log10(N) ) )
        data1 = np.loadtxt( fname1 )
        reg1 = data1[ :, 1 ] > 0.0
        data1 = data1[ reg1, : ]

        attenuation = 0.0002
        fname2 = ( 'entRate(%satt=%.3fdBkm,cohTime=%s%smemEff=%.2f,detEff=%.2f,N=1E%d).txt' 
                 % ( 'stoch,' if stochastic and coherence_time_avg > 0.0 and coherence_time_stdev > 0
                     else '',
                     attenuation * 1000,
                     '%.3fms' % ( coherence_time_avg / 1e-3 ) if coherence_time_avg > 0.0 else 'inf',
                     ( '+-%.2fmcs,' % ( coherence_time_stdev / 1e-6 )
                       if ( stochastic and coherence_time_avg > 0 and coherence_time_stdev > 0 )
                       else ',' ),
                     mem_eff, det_eff, np.log10(N) ) )
        data2 = np.loadtxt( fname2 )
        reg2 = data2[ :, 1 ] > 0.0
        data2 = data2[ reg2, : ]

        coherence_time_avg = 0.1e-3
        attenuation = 0.0
        fname3 = ( 'entRate(%satt=%.3fdBkm,cohTime=%s%smemEff=%.2f,detEff=%.2f,N=1E%d).txt' 
                 % ( 'stoch,' if stochastic and coherence_time_avg > 0.0 and coherence_time_stdev > 0
                     else '',
                     attenuation * 1000,
                     '%.3fms' % ( coherence_time_avg / 1e-3 ) if coherence_time_avg > 0.0 else 'inf',
                     ( '+-%.2fmcs,' % ( coherence_time_stdev / 1e-6 )
                       if ( stochastic and coherence_time_avg > 0 and coherence_time_stdev > 0 )
                       else ',' ),
                     mem_eff, det_eff, np.log10(N) ) )
        data3 = np.loadtxt( fname3 )
        reg3 = data3[ :, 1 ] > 0.0
        data3 = data3[ reg3, : ]

        attenuation = 0.0002
        fname4 = ( 'entRate(%satt=%.3fdBkm,cohTime=%s%smemEff=%.2f,detEff=%.2f,N=1E%d).txt' 
                 % ( 'stoch,' if stochastic and coherence_time_avg > 0.0 and coherence_time_stdev > 0
                     else '',
                     attenuation * 1000,
                     '%.3fms' % ( coherence_time_avg / 1e-3 ) if coherence_time_avg > 0.0 else 'inf',
                     ( '+-%.2fmcs,' % ( coherence_time_stdev / 1e-6 )
                       if ( stochastic and coherence_time_avg > 0 and coherence_time_stdev > 0 )
                       else ',' ),
                     mem_eff, det_eff, np.log10(N) ) )
        data4 = np.loadtxt( fname4 )
        reg4 = data4[ :, 1 ] > 0.0
        data4 = data4[ reg4, : ]


        plt.clf()
        ax = plt.gca()
        plt.loglog( data1[ :, 0 ]/1000, data1[ :, 1 ], '.', ms = 20,#mec = 'none',
                    label = r'$\alpha$ = 0, $t_\mathrm{coh}$ = $\infty$' )
        plt.loglog( data2[ :, 0 ]/1000, data2[ :, 1 ], '.', ms = 20,#mec = 'none',
                    label = r'$\alpha$ = 0.2 dB/km, $t_\mathrm{coh}$ = $\infty$' )
        
        plt.loglog( data3[ :, 0 ]/1000, data3[ :, 1 ], '.', ms = 10,#mec = 'none',
                      label = r'$\alpha$ = %.1f, $t_\mathrm{coh}$ = %s' % (
                          0.0,
                          r'%s%s' % ( '%.0f' % ( coherence_time_avg / 1e-6 )
                                     if coherence_time_avg > 0 else '\infty',
                                     ( ' $\pm$ %.0f mcs' % ( coherence_time_stdev / 1e-6 )
                                       if ( stochastic and coherence_time_avg > 0 )
                                       else ' mcs' ) ) ) )
        plt.loglog( data4[ :, 0 ]/1000, data4[ :, 1 ], '.', ms = 10,#mec = 'none',
                      label = r'$\alpha$ = %.1f dB/km, $t_\mathrm{coh}$ = %s' % (
                          0.2,
                          r'%s%s' % ( '%.0f' % ( coherence_time_avg / 1e-6 )
                                     if coherence_time_avg > 0 else '\infty',
                                     ( ' $\pm$ %.0f mcs' % ( coherence_time_stdev / 1e-6 )
                                       if ( stochastic and coherence_time_avg > 0 )
                                       else ' mcs' ) ) ) )
        
        plt.loglog( [2*2e5*coherence_time_avg,]*2, [1, 2.5e4], 'k--', lw = 1.5 )
        plt.legend( loc = 3, fontsize = 16, frameon = False, numpoints = 1 )
        plt.xlim( 1, 102 )
        plt.ylim( 1, 2.5e4 )

        ax.tick_params( axis='both', which='major', labelsize=13 )
        ax.tick_params( axis='both', which='minor', labelsize=11 )
        
        plt.xlabel( 'Distance, km', fontsize = 15 )
        plt.ylabel( 'Entanglement gen. rate, 1/sec', fontsize = 15 )
        plt.savefig(
            ( 'entRate(att=%.3fdBkm,cohTime=%s%smemEff=%.2f,detEff=%.2f,N=1E%d).pdf' 
                 % ( 0.2,
                     '%.3fms' % ( coherence_time_avg / 1e-3 ) if coherence_time_avg > 0.0 else 'inf',
                     ( '+-%.2fmcs,' % ( coherence_time_stdev / 1e-6 )
                       if ( stochastic and coherence_time_avg > 0  and coherence_time_stdev > 0  )
                       else ',' ),
                     mem_eff, det_eff, np.log10(N) ) )
            )
        plt.show()

```

xwu64:
Thanks for your contribution!

I have some comments about current version of code. 

1. Could you describe the distribution of random coherence time in the docstring of class? Could you try to refer the paper/web of the model in the docstring if it's possible?
2. Because the function `states.truncnorm.rvs` uses the scipy library, you may want to add scipy library to the requirements and setup of the library. 
3. Could you let people set the seed of function `states.truncnorm.rvs`? We may want to use seed to guarantee the reproducibility of simulations.
4. It seems you did not check the distribution of random coherence time in the unit test. Do you think it's necessary or useful to check the distribution in the unit test? (I'm not sure how to check the distribution) Maybe it's fine to ignore such check.

viachesl:
> Thanks for your contribution!
You're welcome.

> 1. Could you describe the distribution of random coherence time in the docstring of class? Could you try to refer the paper/web of the model in the docstring if it's possible?
> 2. Because the function `states.truncnorm.rvs` uses the scipy library, you may want to add scipy library to the requirements and setup of the library.
Sorry, I'm not familiar with documentation generation you use. I've just added some strings in comments describing the extra input variable added. If you send me some guide or link how to do this, I could try.

> 3. Could you let people set the seed of function `states.truncnorm.rvs`? We may want to use seed to guarantee the reproducibility of simulations.
This function uses numpy's seed, I've demonstrated this in test function that is also updated in this commit.

> 4. It seems you did not check the distribution of random coherence time in the unit test. Do you think it's necessary or useful to check the distribution in the unit test? (I'm not sure how to check the distribution) Maybe it's fine to ignore such check.
The test function I've mentioned above does a simple check of the distribution. I calculates and checks average and standard deviation of the distribution. I'm also not sure that further checks are needed. The code in the previous message generates and plots data showing that the distribution works fine.
entRate(att=0 200dBkm,cohTime=0 100ms+-20 00mcs,memEff=0 90,detEff=0 80,N=1E3)

xwu64:
1. You could add description in the comment.

https://github.com/sequence-toolbox/SeQUeNCe/blob/c1eeb2a60c5df2d962933c3bb1c44c12d6e422c8/src/components/memory.py#L300

The documentation could be generated automatically by the `sphinx`. The tutorial of generating documentation could be found at [here](https://github.com/sequence-toolbox/SeQUeNCe/tree/master/docs). 

I think we could merge your pull request once the description is completed.  

Thank you!

PR #73, opened 2021-04-28
fix(reservation.py): Fix refactor

LesterKim:
### Description

- Fixed refactor file for reservation protocol (for creating new methods and variables)

### Stakeholder(s)
@Alex-Kolar
@xwu64 
@acbarnett 
@kinetic-cipher
@nikoaliro

### DIFF
--- a/src/network_management/reservation.py
+++ b/src/network_management/reservation.py
@@ -283,7 +283,7 @@ class ResourceReservationProtocol(StackProtocol):
                 def requirement(protocols):
                     new_protocols = self._initialize_ep_rule_action_requirement(protocols, memories_info)
 
-                    if len(new_protocols) != 2:
+                    if len(new_protocols) < 2:
                         return None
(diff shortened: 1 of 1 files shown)


xwu64:
Thanks for your contribution first. 

We do refactor this file on [the other branch](https://github.com/sequence-toolbox/SeQUeNCe/blob/parallel/src/network_management/reservation.py). This implementation move functions of actions and conditions outside the class. We think it's similar as what you did in this change (make them as the static function). 

Although these changes are not merged into the master branch, we hope our existing work could save your time about refactoring. Could you check our code and share your feedback? 

Apart from that, I agree changes about variable names and coding style.

caitaozhan:
This PR will not be merged because the current code base has changed a lot compared to 3 years ago.

But I have incorporated some refactor ideas from this PR and has been implemented in PR #182

PR #107, opened 2021-10-01
Make components use provided prng

xwu64:


xwu64:
> Hi Xiaoliang, I think the codes themselves have no problem, so this pull request's basic idea is to involve generators to use independent random number streams, and then we need to take care of entities' owners? If this is the case I think it might be even better to add some docstrings to entity, quantum_state, etc. to explain the necessity of these prng's.

I add some docstrings to entity and quantum_state. Please check.

PR #113, opened 2021-11-11
change the type of channel receiver to string

xwu64:
1. Change the receiver of quantum channel and classical channel to string instead of object. This change is to enable parallel simulation merged in the future.
2. Update related code in example and tutorial
3. Refactor unit test code of purification and generation protocol to remove redundant code


Alex-Kolar (approved):
In test_topology.py, I don't think "NotImplementError" is the best here; perhaps "Logic Error" or similar

Alex-Kolar (approved):

PR #117, opened 2021-11-18
Merge RnD to master branch

xwu64:
@Alex-Kolar Could you update version and release?


Alex-Kolar (approved):
Some of the components did not have proper support for the new node generator, which was causing poor performance for the QKD example files. I've updated the components in question and the branch should be ready to merge once the automated test finishes running. I can make the merge.

PR #119, opened 2021-11-19
[Generation] negotiate emit time by two messages

xwu64:


Alex-Kolar (approved):
Looks like it will perform properly. However, after reviewing the code, it seems like we might make a redundant call to set the memory state to the + state (we do this in the start and update_memory methods). I think the call in start might be removed.

xwu64:
> Looks like it will perform properly. However, after reviewing the code, it seems like we might make a redundant call to set the memory state to the + state (we do this in the start and update_memory methods). I think the call in start might be removed.

Which lines do you suggest to remove? The generation class has two place to set state to + state. One is in the function `update_memory` (line 216). One is in the function `emit_event` (line 258). Should we remove line 216?

Alex-Kolar:
Yes, I was thinking of removing line 216. That way, if we add some decoherence decay for the + state in the future, we will minimize the time between activating this state and emitting photons.

xwu64:
> Yes, I was thinking of removing line 216. That way, if we add some decoherence decay for the + state in the future, we will minimize the time between activating this state and emitting photons.

fixed. Please review again.

Alex-Kolar (approved):

PR #123, opened 2021-12-15
Source directory parallelization

Alex-Kolar:
Add parallelized timelines; update some components; update unit tests


Alex-Kolar:
It seems that commit c8fbaba overlaps with pull request #122; both pull requests should produce the same result

xwu64:
> It seems that commit [c8fbaba](https://github.com/sequence-toolbox/SeQUeNCe/commit/c8fbaba61ce248f3c7964ed5619c7837f5028159) overlaps with pull request #122; both pull requests should produce the same result

We could close request #122.

xwu64 (approved):

PR #131, opened 2022-01-07
Examples

Alex-Kolar:
Updating the parallel examples to work with current code; removing further async references


parallel/example/linear_topo.py
```diff
@@ -28,24 +28,24 @@ def main(config_file: str, src: str, dst: str, start_t: int, end_t: int,
...
-                q_links[_dst] = [_src]
-
-        for bsm_node, router_pair in q_links.items():
-            n1, n2 = router_pair
-            if tl.foreign_entities.get(n1, tl.id) != tl.foreign_entities.get(
-                    n2, tl.id):
-                tl.async_entities.add(bsm_node)
-                if bsm_node in tl.entities:
-                    tl.move_entity_to_async_tl(bsm_node)
+    # with open(config_file, 'r') as fh:
```
xwu64: why not delete it directly?

parallel/example/linear_topo.py
```diff
@@ -85,19 +85,14 @@ def main(config_file: str, src: str, dst: str, start_t: int, end_t: int,
...
-    #         msg_type]
 
     with open('%s/linear_perf_%d.json' % (log_path, tl.id), 'w') as fh:
         dump(perf_info, fh)
 
 
 if __name__ == "__main__":
     main("linear_32.json", "router_0", "router_31", 100e12, 100.1e12,
-         50, 0.9, "old/")
+         50, 0.9, "old")
```
xwu64: change log_path "old" to a more reasonable word.

parallel/example/net_with_flow_info.py
```diff
@@ -2,7 +2,7 @@
 from time import time
 import pandas as pd
 from mpi4py import MPI
-from json import load, dump
+from json import dump
```
xwu64: why not change code at [here](https://github.com/sequence-toolbox/SeQUeNCe/blob/f8c326e6aebb98f92f5de60e345b0f5da9463982/parallel/example/net_with_flow_info.py#L33)?

xwu64 (commented):
It's  great to use `add_default_args`. It seems not all functions in the "generator_utils.py" is used in json_config_generators. Should we just import necessary functions?

Alex-Kolar:
I addressed some of your comments:

- Yes, the commented async code in linear_topo can be removed; I removed it
- I changed the linear_topo log path to `"log"`
- Only necessary generator_utils functions are imported to each json config file
- `json.load` has been added back in. I think my IDE had noticed the import from `RouterNetTopo` and marked load as duplicate, but I think it is good to write it explicitly here.

xwu64 (approved):
You can merge this branch after you fix conflict

PR #133, opened 2022-01-11
fix inconsistent formalism used in the ParallelTimeline

xwu64:
use `KET_STATE_FORMALISM` instead of raw string to represent formalism used in the ParallelTimeline


Alex-Kolar (commented):
I noticed this bug too, and fixed it in commit fe4587156ea27a31bc5dba21e6af1d7cd214da76 included in the Examples pull request; the quantum manager client also needed to be fixed. I suggest we only include one commit. I'll request your review on that pull request.

PR #135, opened 2022-01-12
Add unit tests for ParallelTimeline

xwu64:


Alex-Kolar (approved):

Alex-Kolar:
Should we add the parallel tests to the workflow?

xwu64:
> Should we add the parallel tests to the workflow?

I think so. Let me try.

Alex-Kolar (approved):

PR #152, opened 2022-10-23
Align tutorial chapter1 snippets to des_example1.py

zigen:
while I was trying [the tutorial](https://sequence-toolbox.github.io/tutorial/chapter1/discrete-event-simulation.html), I faced a problem. and this PR fixes it.

in chapter1, `store.open` exists as an attribute and a method and conflicts in the tutorial code snippets.
des_example1.py distinguish these as `store.open()` and `store.opening`. so I aligned with it.


Alex-Kolar:
Thanks for editing this, I've merged the commit. In the future, if you make more pull requests, I would suggest setting the target branch as `RnD`; this branch is where we generally integrate and test changes before release.

PR #189, opened 2024-06-27
Small modification of ./docs/source/conf.py for MacOS users.

kim2167:
OS: MacOS _ Sonoma 14.5

Issue: Error and additional warnings are generated due to DS_Store file which is generated by MacOS (It is used to store custom attributes of its containing folder, such as folder view options, icon positions, and other visual information.).

Solution: Modification of line 46.


caitaozhan:
Thanks.

PR #192, opened 2024-08-16
Version 0.6.3

caitaozhan:
Various updates for version 0.6.3

### Added
- We have an Icon for SeQUeNCe!
- `pip install sequence` is now available!
- `src` folder is renamed to `sequence`
- Add `sequence/constants.py` to organize the common constants
- Add support for Python 3.12
- Add `pyproject.toml`

### Changed
- Various refactoring, include but not limited to updating variable/method names, moving code around, adding/rewriting comments, and making the code more succinct, etc.
- In README.md, installation with option `--editable`  is encouraged.
- `logging` looks better
- In `class Reservation`, added attribute `entanglement_number` and `id`
- Update `ResourceReservationProtocol.load_rules()`
- Move `utils/json_config_generators/generator_utils.py` to `sequence/utils/config_generator.py`

### Removed
- Removed support for Python 3.8
- Removed `setup.py`
- Removed `MANIFEST.in`


.github/workflows/main.yml
Alex-Kolar, line 1: Does mpich work with any other versions than 3.12? Should we remove altogether?

docs/source/tutorial/chapter2/example1.py
Alex-Kolar, line 1: Do we need to add logging?

docs/source/tutorial/chapter6/custom_app.py
Alex-Kolar, line 1: Do we need logging?

sequence/kernel/timeline.py
Alex-Kolar, line 1: Is adding a log for every event too much output? I think this may also affect performance.

sequence/entanglement_management/generation.py
anshsingal, line 1: Calling it the Barret Kok protocol rather than the "double heralded protocol" in the docstring of the start method might be clearer.

sequence/entanglement_management/swapping.py
anshsingal, line 1: When the swapping fails, shouldn't the memories be updated instantaneously, and the nodes know about it only when the message is received? That is, shouldn't the resource_manager be updated instantaneously rather than waiting for the message to be received?

sequence/kernel/timeline.py
anshsingal: If we log for debugging alone, performance loss due to logging may not matter during regular execution (right?). However, such high volumes of data might be difficult to analyze anyway. I found logging all classical communications throughout the simulations alone very helpful.

sequence/network_management/network_manager.py
anshsingal, line 1: Should the `swapping_success_rate`  be hardcoded to 0.5?

sequence/topology/topology.py
```diff
@@ -29,10 +29,10 @@ class Topology(ABC):
         tl (Timeline): the timeline used for simulation
     """
 
-    ALL_CC_CONNECT = "cconnections"
+    ALL_CC_CONNECT = "cconnections"    # NOTE why not name it ALL_C_CONNECT?
```
anshsingal: ALL_ClassicalChannel_CONNECT I believe?

sequence/network_management/network_manager.py
allenyxzang: Swapping is implemented on the repeater node, so BSM on matter memories is possible, which do not have success probability upper bound.

.github/workflows/main.yml
caitaozhan: It should work. And to be honest, I don't know much about mpich. Let you decide on mpich

docs/source/tutorial/chapter2/example1.py
caitaozhan: Since logging never appeared in the tutorial. Let's keep the logging here.

docs/source/tutorial/chapter6/custom_app.py
caitaozhan: I'll remove the logging here

sequence/kernel/timeline.py
caitaozhan: When I started using SeQUeNCe, adding logging to the timeline helped when I was playing around with small examples. Later, I switched off the logging in timeline by not adding 'timeline' to the logging modules, so the impact on performance is minimal.

sequence/entanglement_management/generation.py
caitaozhan: Yes. Updated

sequence/entanglement_management/swapping.py
caitaozhan: Understand your point. I made this change so that the swapping node memory updates at the same time as the end node memory. If these memories are not updated at the same time, there will be a time gap. This time gap in updating the memories at the swapping node and end nodes will create issues for the Adaptive continuous protocol.

sequence/topology/topology.py
caitaozhan: It seems that here a channel is single-direction, and a connection is bi-direction. So do you think "ALL_ClassicalChannel_CONNECT" is a little strange, because it is saying it is both single-direction and bi-direction?

sequence/entanglement_management/generation.py
caitaozhan: Fixed

sequence/network_management/network_manager.py
caitaozhan: How high can the probability be here? Could it be close to 1? It would be great to have a link or reference.

sequence/entanglement_management/swapping.py
caitaozhan: I have reverted the changes in protocol EntanglementSwappingA

sequence/network_management/network_manager.py
allenyxzang: When considering quantum circuits on matter-based systems we usually do not have the notion of success probability. This is because the measurement can always have outcome, which is different from when using intrinsic degrees of freedom of photons that can be lost. Therefore, errors in quantum circuits on matter-based systems are mainly captured by operation infidelities.

PR #204, opened 2025-01-03
[major] Add New Module: Quantum Transducer related

caitaozhan:
New Component Modules from this paper: https://arxiv.org/pdf/2411.11377
- `Transducer`
- `Transmon`
- `FockDetector`
- `FockBeamSplitter2`

Also,
- Added the unit tests in the folder `tests`
- Added examples in the folder `example/QuantumTransduction`
- Reorganized the `AbsorptiveMemory` related files
- Update the `requirements.txt`
- Update `constants` module
- Update comments here and there

NOTE: work in progress. 
- Quantum Channel is not used
- Scheduling events all at once and pass in class as argument instead of object


Alex-Kolar (commented):
Code looks good. A couple other minor things I forgot to mention in the meeting this morning:

1. Do we want to rename the `QuantumTransduction` and `AbsorptiveMemory` folders (and subfiles) to be snake case instead of camel case? This would match the rest of the project
2. I don't remember if I mentioned this previously but it looks like the badges on the README/main github page are off center. Is it possible to fix this? I attached a photo.
Screenshot 2025-01-17 at 12 09 29 PM

If these don't get changed, it's no big deal, just though I'd bring them up.

caitaozhan:
@Alex-Kolar 

image

I moved the badges to the center, but there are some annoying "_" between the badges. I tried various ways, but I can not remove them. 

Actually, the annoying "_" is the reason why the badges are previous on the left side, instead of being center. I failed to make it center without "_". The only way I removed the "_" is by using the Markdown syntax, but that led the badges to be on the side.

PR #219, opened 2025-06-18
[major] noise.py and test_noise.py

SagarPatange:
- added `Noise` class which has methods that simulate depolarization and measurement noise
- added 3 new test cases in `test_noise.py`


caitaozhan:
Hi Sagar,

Thanks! It looks good.

Best,
Caitao

PR #227, opened 2025-08-06
226 fix protocol delay

hayekr:
Brings over changes from quisp collab branch

### DIFF
--- a/sequence/components/bsm.py
+++ b/sequence/components/bsm.py
@@ -147,6 +147,8 @@ class BSM(Entity):
                            (complex(0), complex(sqrt(1 / 2)), complex(sqrt(1 / 2)), complex(0)),
                            (complex(0), complex(sqrt(1 / 2)), -complex(sqrt(1 / 2)), complex(0)))
 
+        self.rng = self.get_generator()
+
     def init(self):
         """Implementation of Entity interface (see base class)."""
 
@@ -251,7 +253,7 @@ class PolarizationBSM(BSM):
         self.photons[0].combine_state(self.photons[1])
 
         # measure in bell basis
-        res = Photon.measure_multiple(self.bell_basis, self.photons, self.get_generator())
+        res = Photon.measure_multiple(self.bell_basis, self.photons, self.rng)
 
         # check if we've measured as Phi+ or Phi-; these cannot be measured by the BSM
         if res == 0 or res == 1:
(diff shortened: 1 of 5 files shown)


sequence/components/bsm.py
caitaozhan, line 1: `get_generator()` is used all over the place in SeQUeNCe. Doing `self.rng` in this file will bring inconsistency.

I don't see a good reason to break the convention, i.e., `get_generator()`. Saving a few characters? 

Doing `get_generator()` is probably written by @Alex-Kolar, what are your thoughts?

sequence/resource_management/resource_manager.py
caitaozhan, line 1: If you'd like to accommodate this limited case, please do so in your customized class (make a subclass in your project, and do whatever you want there)

Please do not change the original class.

Also, putting the `delay` in the `resource_manager.py` doesn't look good to me. Since you are accommodating the BSM in routers, putting the delay in the `swapping protocol` may look better.

sequence/entanglement_management/generation.py
caitaozhan, line 1: The delay looks good. Let's put the good-looking figures that Ansh created into the tutorial.

sequence/components/bsm.py
Alex-Kolar: Yes, this is used all over the code base. It looks like there was a bug that was related to this usage with the BSM. I'll have to investigate more to see if this bug is also present in other areas of the codebase, and if it is, we should probably change this line everywhere.

Alex-Kolar:
I think the failed tests are due to a missing import. I can push a fix to this.

anshsingal:
I made a small change to the optical_channel.py file (in schedule_transmit) where the previous calculation was leading to numerical errors (simple rearrangement of brackets). Everything else remains the same. The failed commit was because of a wrong change made to the BSM file which has been undone in the latest commit.

anshsingal:
For large numbers, the original expression produces numerical artifacts whereas the new one does not. Its a limitation with floating point arithmetic because when we multiply min_time and self.frequency, that is an integer of the order 1e20 or so, where floating point arithmetic starts producing significant errors. 

image

anshsingal:
Although, I can see that for some other inputs, we get the opposite issue as well. So, not sure which fix to use in this case. 

image

caitaozhan:
After many trial and error, I think the floating issue is finally solved LOL

```python
import gmpy2
gmpy2.get_context().precision = 200
```

```python
    def schedule_transmit(self, min_time: int) -> int:
       
        min_time = max(min_time, self.timeline.now())
        # use the package gmpy2 for high precision floating point, set to 200 bits (see beginning of this file)
        min_time = gmpy2.mpfr(min_time)
        freq = gmpy2.mpfr(self.frequency)
        second = gmpy2.mpfr(SECOND)
        time_bin = min_time * freq / second

        if time_bin - gmpy2.floor(time_bin) > gmpy2.mpfr(EPSILON):
            time_bin = int(time_bin) + 1       # round to the next time bin
        else:
            time_bin = int(time_bin)

        # find earliest available time bin
        while time_bin in self.send_bins:
            time_bin += 1
        hq.heappush(self.send_bins, time_bin)

        # calculate time
        time = int(time_bin * (1e12 / self.frequency))
        return time

```

PR #229, opened 2025-08-11
Fix long-lasting floating point issues

caitaozhan:
This PR is mostly for fixing a long-lasting floating-point issue in `QuantumChannel.schedule_transmit()` and `QuantumChannel.transmit()`, which causes `assert emit_time == msg.emit_time` in `EntanglementGenerationA.received_message()` to throw an error. I encountered this problem in 2024, but it later mysteriously disappeared LOL. So I didn't bother it. All of a sudden, the problem resurfaces this summer LOL.

The key to fixing the floating-point problem is to understand that the **float in Python is 64 bits, which is equivalent to around 16 significant decimal digits.** Solution: use package **gmpy2** and **set precision to 200 bits, which is equivalent to around 54 significant decimal digits.**

To make the code more elegant, two new methods are introduced: `QuantumChannel.time_to_timebin()` and `QuantumChannel.timebin_to_time()`

Other minor updates in this PR:
1. Rename `QuantumNode` to `DQCNode`, which is a node designed for distributing quantum computing
2. Add a new method `NetworkManager.update_forwarding_table()`
3. Several cosmetic updates

### DIFF
--- a/sequence/components/bsm.py
+++ b/sequence/components/bsm.py
@@ -147,6 +147,7 @@ class BSM(Entity):
                            (complex(0), complex(sqrt(1 / 2)), complex(sqrt(1 / 2)), complex(0)),
                            (complex(0), complex(sqrt(1 / 2)), -complex(sqrt(1 / 2)), complex(0)))
 
+
     def init(self):
         """Implementation of Entity interface (see base class)."""
 
--- a/sequence/entanglement_management/generation.py
+++ b/sequence/entanglement_management/generation.py
@@ -5,7 +5,7 @@ Also defined is the message type used by this implementation.
 Entanglement generation is asymmetric:
 
 * EntanglementGenerationA should be used on the QuantumRouter (with one node set as the primary) and should be started via the "start" method
-* EntanglementGeneraitonB should be used on the BSMNode and does not need to be started
+* EntanglementGenerationB should be used on the BSMNode and does not need to be started
 """
 
 from __future__ import annotations
@@ -271,11 +271,10 @@ class EntanglementGenerationA(EntanglementProtocol):
             other_qc_delay = msg.qc_delay
             self.qc_delay = self.owner.qchannels[self.middle].delay
             cc_delay = int(self.owner.cchannels[src].delay)
-            total_quantum_delay = max(self.qc_delay, other_qc_delay)  # two qc_delays are the same for "meet_in_the_middle"
 
             # get time for first excite event
             memory_excite_time = self.memory.next_excite_time
-            min_time = max(self.owner.timeline.now(), memory_excite_time) + total_quantum_delay - self.qc_delay + cc_delay  # cc_delay time for NEGOTIATE_ACK
+            min_time = max(self.owner.timeline.now(), memory_excite_time) + other_qc_delay - self.qc_delay + cc_delay  # cc_delay time for NEGOTIATE_ACK
             emit_time = self.owner.schedule_qubit(self.middle, min_time)  # used to send memory
             self.expected_time = emit_time + self.qc_delay  # expected time for middle BSM node to receive the photon
 
--- a/sequence/kernel/quantum_manager.py
+++ b/sequence/kernel/quantum_manager.py
@@ -6,9 +6,9 @@ The states may currently be defined in two possible ways:
 
 The manager defines an API for interacting with quantum states.
 """
-
 from __future__ import annotations
-from abc import abstractmethod
+
+from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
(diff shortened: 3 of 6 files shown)


sequence/components/optical_channel.py
```diff
@@ -6,10 +6,10 @@
 """
 
 import heapq as hq
+import gmpy2
+gmpy2.get_context().precision = 200
```
hayekr: Formatting: this should not be with imports

sequence/components/optical_channel.py
```diff
@@ -184,6 +184,46 @@ def transmit(self, qubit: "Photon", source: "Node") -> None:
         else:
             pass
 
+    def time_to_timebin(self, time: int, frequency: float) -> int:
```
hayekr, line 191: I think that time_to_timebin and timebin_to_time can be consolidated to one function.

sequence/components/optical_channel.py
```diff
@@ -184,6 +184,46 @@ def transmit(self, qubit: "Photon", source: "Node") -> None:
...
+
+        Args:
+            time (int): simulation time to convert.
+            frequency (float): frequency of the channel.
+        Returns:
+            int: time bin corresponding to the given simulation time.
+        """
+        time = gmpy2.mpfr(time)
+        frequency = gmpy2.mpfr(frequency)
+        ps_per_second = gmpy2.mpfr(SECOND)
```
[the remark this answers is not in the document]
hayekr: I agree with the robot

sequence/components/optical_channel.py
```diff
@@ -184,6 +184,46 @@ def transmit(self, qubit: "Photon", source: "Node") -> None:
...
+        time_bin = time * frequency / ps_per_second
+        if time_bin - gmpy2.floor(time_bin) > gmpy2.mpfr(EPSILON):
```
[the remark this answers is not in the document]
hayekr: I also agree with this.

sequence/constants.py
```diff
@@ -20,14 +20,16 @@
 EPSILON: Final = 1e-8
 
 # convert to picosecond
-NANOSECOND:  Final = 1e3
-MICROSECOND: Final = 1e6
-MILLISECOND: Final = 1e9
-SECOND:      Final = 1e12
+NANOSECOND:  Final = 10**3
```
hayekr, line 23: Is the intention of this to make these constants type:int instead of floats

sequence/constants.py
```diff
@@ -20,14 +20,16 @@
...
-SECONDS_PER_MINUTE = MINUTES_PER_HOUR = 60
-CARRIAGE_RETURN = '\r'
-SLEEP_SECONDS = 3
+NANOSECONDS_PER_MILLISECOND: Final = 10**6
+PICOSECONDS_PER_NANOSECOND:  Final = 10**3
+NANOSECONDS_PER_MICROSECOND: Final = 10**3
+MILLISECONDS_PER_SECOND:     Final = 10**3
+SECONDS_PER_MINUTE: Final = 60
+MINUTES_PER_HOUR:   Final = 60
+CARRIAGE_RETURN:    Final = '\r'
```
hayekr: Why should we store this?

CHANGELOG.md
```diff
@@ -6,6 +6,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [0.7.3] - 2025-8-1
+
+### Added
+- new class `QuantumNode` under _topology.node_. Adds QuantumNode class extending QuantumRouter with data memory and teleportation capabilities
```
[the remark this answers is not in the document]
hayekr: This is correct.

sequence/components/optical_channel.py
caitaozhan: Where do you suggest putting it?

CHANGELOG.md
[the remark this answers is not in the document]
caitaozhan: That changelog is for version 0.7.3. In verison 0.7.3, the name is indeed `QuantumNode`

sequence/components/optical_channel.py
hayekr: After imports, first line after

sequence/components/optical_channel.py
caitaozhan: I don't think that is a good idea. Looks bad to me.

sequence/components/optical_channel.py
caitaozhan: Sounds good

sequence/constants.py
caitaozhan: These are legacy code... They should be used somewhere in the codebase.

sequence/constants.py
caitaozhan: Yes. `1e3` is a float, but `10**3` is an int.

hayekr: Yes my comment was asking if this was the intention of the change. If not, what is the purpose of this change.

PR #234, opened 2025-08-15
Trajectree integration

anshsingal:
Pull request for the plugin mechanism in SeQUeNCe. For the first version, trying out the integration with Trajectree.


sequence/topology/router_net_topo.py
```diff
@@ -63,10 +68,11 @@ def _load(self, filename: str):
 
     def _add_timeline(self, config: dict):
         stop_time = config.get(Topo.STOP_TIME, float('inf'))
+        trunc = config.get(Topo.TRUNC, float(4))
```
caitaozhan: Why the default number is 4? Will 1 be more appropriate?
trunc = 1 means dimension = 2, i.e., qubits, the most common situation.

Also, should it be int instead of float?

sequence/topology/node.py
```diff
@@ -23,12 +23,21 @@
...
 from ..components.detector import QSDetector, QSDetectorPolarization, QSDetectorTimeBin
 from ..qkd.BB84 import BB84
 from ..qkd.cascade import Cascade
-from ..entanglement_management.generation import EntanglementGenerationB
+# from ..entanglement_management.generation import EntanglementGenerationB
+
+from importlib import import_module
+from ..config import CONFIG
+import_formalism = CONFIG.get('formalism', "")
+try:
```
caitaozhan: Please add some comments here to explain what is happening.

sequence/network_management/reservation.py
```diff
@@ -13,8 +13,15 @@
     from ..resource_management.memory_manager import MemoryInfo, MemoryManager
     from ..entanglement_management.entanglement_protocol import EntanglementProtocol
 
+from importlib import import_module
```
caitaozhan, line 16: Please add some comments here to explain what is happening.

sequence/entanglement_management/generation_trajectree.py
```diff
@@ -0,0 +1,316 @@
...
+            # configure params
+            # Note again, that all of these corresponf to the delays to the BSM node.
+            other_qc_delay = msg.qc_delay
+            other_cc_delay = msg.cc_delay
+            total_quantum_delay = max(self.qc_delay, other_qc_delay)
+            total_classical_delay = max(self.cc_delay, other_cc_delay)
+
+            end_node_cc_delay = self.owner.cchannels[self.remote_node_name].delay
+
+            # print("delays: self.qc_delay, other_qc_delay, total_quantum_delay, self.cc_delay, other_cc_delay, total_classical_delay", self.qc_delay, other_qc_delay, total_quantum_delay, self.cc_delay, other_cc_delay, total_classical_delay)
```
caitaozhan: Please remove those commented print(). 

Some print() could be replaced with `log.logger.debug()`

sequence/entanglement_management/generation.py
```diff
@@ -26,7 +26,7 @@
 from ..components.circuit import Circuit
 from ..utils import log
 
-
+print("Using default backend for entanglement generation")
```
caitaozhan: Please replace with `log.logger.info`. Do not use print() anywhere; use the logging system.

sequence/topology/router_net_topo.py
anshsingal: So, the truncation in such bosonic systems is for the number of photons in the modes which form the bi-mode system (which we use as our "qubits" in photonic communication). Having a truncation of 1 is a more ideal case where you have either 0 or 1 photons in the system, thereby ignoring multi-pair effects entirely. We could set the default trunc to 1 though. That would not cause any issues.

sequence/network_management/reservation.py
anshsingal: Done

caitaozhan:
Looks good to me! Asking Copilot for a final review.

sequence/entanglement_management/generation_trajectree.py
```diff
@@ -0,0 +1,316 @@
...
+from ..utils import log
+
+from functools import lru_cache
+
+import numpy as np
+from trajectree.sequence.swap import perform_swapping_simulation # type: ignore
+
+from ..config import CONFIG
+
+print("Using Trajectree backend for entanglement generation")
```
hayekr: Use log here

sequence/topology/node.py
```diff
@@ -254,8 +271,8 @@ def __init__(self, name: str, timeline: "Timeline", other_nodes: list[str],
             bsm_args = component_templates.get("SingleHeraldedBSM", {})
             bsm = SingleHeraldedBSM(bsm_name, timeline, **bsm_args)
         else:
-            raise ValueError(f'Encoding type {self.encoding_type} not supported')
-
+            bsm = ShellBSM(bsm_name, timeline)
```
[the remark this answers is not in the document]
anshsingal: @caitaozhan, what do you suggest here?

sequence/components/bsm.py
```diff
@@ -708,3 +708,13 @@ def trigger(self, detector: Detector, info: dict[str, Any]):
         info = {'entity': 'BSM', 'info_type': 'BSM_res', 'res': res, 'time': time}
         self.notify(info)
 
+class ShellBSM(BSM):
+    def __init__(self, name, timeline):
+        super(BSM, self).__init__(name, timeline)
```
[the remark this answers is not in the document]
anshsingal: The constructor here is meant to skip to BSM's parent and not implement BSM's constructor (they need some additional entities not required in the shellBSM
)

sequence/topology/node.py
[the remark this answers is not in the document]
anshsingal: The default is kept as such since the simulator has no way of knowing if there is a new formalism being used already. Hence, raising an error is not the best idea here.

sequence/components/bsm.py
```diff
@@ -708,3 +708,13 @@ def trigger(self, detector: Detector, info: dict[str, Any]):
...
+        self.encoding = "shell"
+
+    def get(self, photon, **kwargs):
+        pass
+
+    def trigger(self, detector: Detector, info: dict[str, Any]):
```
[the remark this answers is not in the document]
anshsingal: Added Docstrings

sequence/topology/node.py
```diff
@@ -67,6 +83,7 @@ def __init__(self, name: str, timeline: "Timeline", seed=None, gate_fid: float =
         self.owner = self
         self.cchannels = {}  # mapping of destination node names to classical channels
         self.qchannels = {}  # mapping of destination node names to quantum channels
+        self.cached_entanglement = {}
```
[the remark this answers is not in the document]
anshsingal: Removed this line and added this in the application layer.

PR #236, opened 2025-08-20
Renaming some classes related to Distributed Quantum Computing, etc.

caitaozhan:
- Rename `QuantumNodeNetTopo` to `DQCNetTopo`
- Minor updates inside class `DQCNode`
- Update docstrings
- Add new method `Node.get_component_by_name()`
- The `RouterNetTopo` now reads in the formalism of the quantum state in the json config file.
- Update the related pytests


ssuresh27-byte:
Looks good to me!

PR #239, opened 2025-08-25
231 update purification and generation

hayekr:
This PR has many significant changes.

1. Define new ABC for generation and purification from the large class files.
2. Create class methods to follow the registry design pattern
3. Implement a global typing that can be used to set the formalism/type on the user side
4. Implement decorator registration with base classes to allow users to implement their own protocols and easily switch sequence over to the newly made/experimental protocol.

The new classes are exposed and therefore made available in the same manner as the previous implementation.


sequence/entanglement_management/generation/generation.py
anshsingal, line 1: Does this work as an ABC for the other generation protocols to be implemented?

sequence/kernel/timeline.py
```diff
@@ -54,7 +55,7 @@ class Timeline:
         show_progress (bool): show/hide the progress bar of simulation.
         quantum_manager (QuantumManager): quantum state manager.
     """
-    def __init__(self, stop_time: int = inf, formalism = KET_STATE_FORMALISM, truncation = 1):
+    def __init__(self, stop_time: int = inf, formalism: str = '', truncation = 1):
```
anshsingal, line 58: Without a default, what is chosen by the simulator by default and where is that decision made?

sequence/network_management/reservation.py
```diff
@@ -121,28 +120,30 @@ def eg_rule_condition(memory_info: "MemoryInfo", manager: "MemoryManager", args:
 
 # entanglement purification
 
-def ep_rule_action1(memories_info: list["MemoryInfo"], args: Arguments) -> tuple[BBPSSW, list[str], list["ep_req_func1"], list[dict]]:
+def ep_rule_action1(memories_info: list["MemoryInfo"], args: Arguments) -> tuple[BBPSSWProtocol, list[str], list["ep_req_func1"], list[dict]]:
```
anshsingal, line 121: Should this be hard coded to the BBPSSW protocol, or can that also be configured as a part of the plugin?

hayekr: It should be hardcoded to BBPSSWProtocol, this is the ABC and all plugin's will extend this.

sequence/entanglement_management/generation/generation.py
hayekr: Correct.

sequence/network_management/reservation.py
anshsingal: Can we make the purification protocol further abstracted so that the BBPSSW is also a plugin in the simulator?

hayekr:
Completed all the changes. Ready for merge

sequence/topology/router_net_topo.py
```diff
@@ -71,11 +65,11 @@ def _load(self, filename: str):
     def _add_timeline(self, config: dict):
         stop_time = config.get(Topo.STOP_TIME, float('inf'))
         formalism = config.get(Topo.FORMALISM, KET_STATE_FORMALISM)
```
caitaozhan, line 66: if `truncation` is in the config file, then it cannot be passed into the timeline or Quantum Manager

sequence/constants.py
```diff
@@ -24,3 +24,13 @@
...
 SECOND:      Final = 10**12
+
+# Built-In Formalisms
+KET_STATE_FORMALISM: Final = "ket_vector"
+DENSITY_MATRIX_FORMALISM: Final = "density_matrix"
+FOCK_DENSITY_MATRIX_FORMALISM: Final = "fock_density"
+BELL_DIAGONAL_STATE_FORMALISM: Final = "bell_diagonal"
+
+# Built-In Generation Protocols
+BARRET_KOK: Final = 'single_atom'
```
caitaozhan: The `single_atom` is confusing (for a long time); can it be changed to `barret_kok`? `single_atom` is supposed to be a photon encoding type. @Alex-Kolar.

sequence/topology/router_net_topo.py
hayekr: This was the default code from release.
`QuantumManager` never is passed `truncation`
Should `self.tl = Timeline(stop_time)` be replaced with `self.tl = Timeline()`

sequence/entanglement_management/generation/__init__.py
Alex-Kolar, line 1: Very minor but can we make the filenames in snake case to match the general python style?

hayekr: done

Alex-Kolar (approved):

sequence/kernel/timeline.py
caitaozhan, line 1: `Timeline.step_time` is an `integer`. You changed it to `int | float` before, then I reverted it. Please revert it back to `int`. If `inf` is causing an issue, then please replace `inf` with something else that is `int`.

hayekr: If 'inf' is the default value then `Timeline.stop_time` can indeed be a float. `int | float` accounts for this by allowing/type hinting for both.

caitaozhan: I changed `inf` to `10 ** 23`, approximately 3000 years. That avoids the `float` type

The quote below is from the SeQUeNCe paper. So, `1.1 picoseconds` should not be allowed. `1.1` should be rounded to `1`.
> Design and implementation of a scalable, customizable, discrete-event quantum network simulator, SeQUeNCe, that models the behavior of quantum networks with picosecond precision

PR #251, opened 2025-09-12
249 deprecate parallel sequence

hayekr:
The parallel code is moved to a mirror repo: 
[Github repo for parallel](https://github.com/sequence-toolbox/SeQUeNCe-Parallel), which does not contain gh actions for deployments.


hayekr:
It is now a public repo

PR #272, opened 2025-10-07
Update purification protocol

hayekr:
Purification had the following issues:
* When coherence time is inf, purification will return success with a fidelity higher than the target.
* When coherence time is finite, purification succeeds but may not ever return pairs with at least target fidelity. Further, as coherence time shrinks, purification will succeed but never return pairs with at least target fidelity. When pairs never reach target fidelity, **the responder node will enter infinite loop of purification until memory expiry while initiator will deliver insufficient pairs to the app and be released.**

This PR implements the following changes:
* Add a purification state, when purification succeeds, the memory is transitioned to "PURIFIED". This new state is identical in function to the "ENTANGLED" state, but can only be triggered by BBPSSW. The result is purification will succeed once, and deliver to the application. This could be modified to allow for more complicated behavior of purification w.r.t delivered fidelity.
* Add tracking for scheduled expiration events by tracking the event ID. 
* The RequestApp is updated to allow counting of entanglement if the pairs are purified, but below the target fidelity.
* Adds protocol type to BBPSSW (unused)

### DIFF
--- a/sequence/components/memory.py
+++ b/sequence/components/memory.py
@@ -312,7 +312,7 @@ class Memory(Entity):
             Will notify upper entities of expiration via the `pop` interface.
             Will modify the quantum state of the memory.
         """
-
+        #log.logger.debug(f'Memory {self.name} has expired at time {self.timeline.now()}')
         if self.is_in_application:
             pass
 
@@ -411,6 +411,7 @@ class Memory(Entity):
             self.timeline.remove_event(self.expiration_event)
 
         decay_time = self.timeline.now() + int(self.cutoff_ratio * self.coherence_time * 1e12)
+        #log.logger.debug(f'Memory {self.name} is set to expire at time {decay_time}')
         process = Process(self, "expire", [])
         event = Event(decay_time, process)
         self.timeline.schedule(event)
@@ -425,7 +426,7 @@ class Memory(Entity):
         Args:
             time (int): new expiration time.
         """
-
+        #log.logger.debug(f'Memory expiry is updated')
         time = max(time, self.timeline.now())
         if self.expiration_event is None:
             if time >= self.timeline.now():
--- a/sequence/entanglement_management/purification/bbpssw_bds.py
+++ b/sequence/entanglement_management/purification/bbpssw_bds.py
@@ -53,6 +53,7 @@ class BBPSSW_BDS(BBPSSWProtocol):
 
         self.is_twirled = is_twirled
         self.ep_matched = False
+        self.protocol_type = 'bbpssw_bds'
 
     def start(self) -> None:
         """Method to start entanglement purification.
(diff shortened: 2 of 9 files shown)


sequence/entanglement_management/purification/bbpssw_bds.py
```diff
@@ -122,16 +124,7 @@ def received_message(self, src: str, msg: BBPSSWMessage) -> None:
         Side Effects:
             Will call `update_resource_manager` method.
         """
-
-        # check the status of entanglement
-        if self.meas_memo.entangled_memory['node_id'] is None or self.kept_memo.entangled_memory['node_id'] is None:
-            log.logger.info(f'No entanglement for {self.meas_memo} or {self.kept_memo}.')
-            # when the AC Protocol expires, the purification protocol on the primary node will get removed, but the purification protocol on the non-primary node is still there
-            self.owner.protocols.remove(self)
```
Alex-Kolar, line 130: Why did this code get removed? Is it only for the adaptive continuous protocol?

hayekr: It was not present on the Quisp branch. It was ported from ACP by accident, I am unsure if this is is ACP specific or if it is one of the bug fixes @caitaozhan applied.

sequence/kernel/timeline.py
Alex-Kolar, line 1: Why are we only logging the expire events here?

hayekr: I am adding expiry events in addition to all other timeline logs. I am mapping scheduled expiry to actual expiry event. This doesn't remove any logging.

sequence/entanglement_management/purification/bbpssw_bds.py
caitaozhan: ACP related code should be safe to be removed

sequence/resource_management/memory_manager.py
```diff
@@ -63,6 +63,8 @@ def update(self, memory: "Memory", state: str) -> None:
             info.to_occupied()
         elif state == "ENTANGLED":
             info.to_entangled()
+        elif state == "PURIFIED":
```
Alex-Kolar, line 66: Is this method called somewhere with the argument "PURIFIED"? I couldn't find it in the bbpssw files

hayekr: Resolving this. I missed a few references

sequence/components/memory.py
caitaozhan, line 1: Let's either uncomment the `#log.logger.debug`, or delete the commented line.

hayekr: can you be specific on what "old logic" is broken please

caitaozhan: The condition for swapping is `es_rule_conditionA()`, `es_rule_conditionB1()`, and `es_rule_conditionB2()`. Swapping requires the memory to be in ENTANGLED. If the memory goes to PURIFIED, it seems the conditions will never be satisfied. But in theory, purified memory should be able to get swapped.

JoaquinChung: I see two options:
1. We add an OR PURIFIED to the rule
2. We create a new protocol to swap purified pairs, which will be just a copy of the original and we change the condition on the state of the memory.

hayekr: @caitaozhan @JoaquinChung  Please re-review the code. swapping memory checks if the state is in the list ['ENTANGLED', 'PURIFIED']

The condition for swapping should be met.

sequence/network_management/reservation.py
```diff
@@ -247,29 +279,30 @@ def es_req_func(protocols: list["EntanglementProtocol"], args: Arguments) -> Ent
...
+def es_rule_conditionA(memory_info: "MemoryInfo", memory_manager: "MemoryManager", args: Arguments) -> list[
+    "MemoryInfo"]:
     """Condition function used by EntanglementSwappingA protocol on nodes
     """
     memory_indices = args["memory_indices"]
     left = args["left"]
     right = args["right"]
     fidelity = args["fidelity"]
-    if (memory_info.state == "ENTANGLED"
+    if (memory_info.state in ["ENTANGLED", "PURIFIED"]
```
hayekr: @caitaozhan @JoaquinChung 

As shown here this work was already done before the review.

sequence/components/memory.py
hayekr: @caitaozhan If I am incorrect, please comment on the incorrect lines.

caitaozhan: It looks good!

PR #274, opened 2025-11-06
Enhance add_forwarding_rule for multipath routing

MikaelAziz:
Modified add_forwarding_rule to support multiple next hops with probabilities for multipath routing. I am using this for my personal research project, but it should also be useful for various use cases and more complex network routing protocols. Additionally if this change is being adopted then a few of the tutorials should also be modified to include this.

### DIFF
--- a/sequence/network_management/routing.py
+++ b/sequence/network_management/routing.py
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
 
 from ..message import Message
 from ..protocol import StackProtocol
-
+import random
 
 class StaticRoutingMessage(Message):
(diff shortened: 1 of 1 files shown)


caitaozhan:
Hi Mikael,

I see you changed the `add_forwarding_rule`, which will result in failing some pytests. Please do the following:
1. Revert the change that happened to `add_forwarding_rule`
2. Add your new function `add_forwarding_rule_multihop` (or whatever name you want)
3. Add a line somewhere in the pytest to make sure the unit test covers the newly added function.

Thanks,
Caitao

PR #279, opened 2025-12-10
[minor] fixing a bug in gui.py

caitaozhan:
Fixing a bug in gui.py

The original function `run_server()` is replaced with `run()`


hayekr (changes requested):
Bump the version to 0.8.3, as this will be the next release, and RnD should be one version above master so CD pipeline recognizes RnD as a pre-release

hayekr (approved):

PR #297, opened 2026-02-06
Topology Refactor

asmuley2:
## DRY/YAML/Constants Refactor

  - Config loading (JSON/YAML) and template extraction are now centralized in parent Topology._load(). Each child class implements _build(config), which receives the parsed config dict and defines its own build sequence — e.g. MultihopTopo sets up BSM mappings and forwarding tables, while QlanStarTopo extracts QLAN-specific parameters and wires up protocols. The parent handles *how* configs are loaded and parsed; the children define *what* gets built and in what order via _build().
    - _add_timeline() moved to parent Topology. MultihopTopo overrides it to add quantum manager setup.
    - YAML config support. Topology._load_config() now handles both JSON and YAML files. Support also extended to utils/config_generator.py and sequence/gui/run_gui.py.

  - Added intermediate class MultihopTopo in multihop_topo.py, inherited by multi-hop topologies (RouterNetTopo, DQCNetTopo) for shared structures including _build(), _map_bsm_routers(), _add_bsm_node_to_router(), _add_qconnections(), and _generate_forwarding_table(). QlanStarTopo inherits directly from Topology (note that I tried to create an intermediate star_topo class and proved to be an unnecessary premature abstraction).
    - [minor] Method _calc_cc_delay() in parent Topology class to consolidate similar logic between all topologies.

  - All constants centralized to sequence/constants.py. Entire codebase (topology classes, config generators, GUI, tests, examples, docs/tutorials) updated to import constants from sequence.constants instead of accessing them as class attributes.

## Testing
  - YAML test configs and test_yaml_compat.py to verify YAML configs produce the same results
  as their JSON counterparts.
  - Refactored code regularly tested against existing Topology and Qlan pytests. (update: all pytests pass)

## Ideas
  - Make the repo-level utils folder fully usable, including YAML versions of the generation scripts.
  - Adjust QlanStarTopo configuration scheme to use templates, further concentrating logic into the parent Topology class. See tests/topology/qlan_topo_template_config.json for a prototype.
    - Minor issue: this is organizationally a different .json, so we'd need to either abandon backwards compatibility or provide extra fallback logic for it.
    - Would also eliminate the double loop in _add_nodes: currently clients must be created before orchestrators (orchestrator validates against client memory count at construction), forcing two passes over the node list. With templates (perhaps internal), nodes could be created in any order and wired together after.

### DIFF
--- a/sequence/topology/topology.py
+++ b/sequence/topology/topology.py
@@ -53,7 +53,7 @@ class Topology(ABC, metaclass=_DeprecatedAttrMeta):
     A network may also be generated using an external json file.
 
     Attributes:
-        bsm router-to-map
+        bsm router-to-map (dict): mapping BSM nodes to router nodes.
         nodes (dict[str, list[Node]]): mapping of type of node to a list of same type node.
         qchannels (list[QuantumChannel]): list of quantum channel objects in network.
         cchannels (list[ClassicalChannel]): list of classical channel objects in network.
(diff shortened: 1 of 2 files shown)


sequence/topology/topology.py
```diff
@@ -66,51 +45,103 @@ def __init__(self, conf_file_name: str):
...
+        Args:
+            filename: path to the configuration file
+
+        Returns:
+            Configuration dictionary
+        """
+        with open(filename) as fh:
+            if filename.endswith(('.yaml', '.yml')):
+                return yaml.safe_load(fh)
+            return json.load(fh)
```
hayekr: This would result in an error if the user passes non json or yaml files

sequence/constants.py
hayekr, line 1: You cannot remove these without backwards compatibility. This is a breaking change; i.e moving Topology.NAME without deprecation warnings  will break existing user code. 

You must: Add dep warnings for old usage; keep the old interface

sequence/topology/multihop_topo.py
hayekr, line 1: The mixed inheritance pattern, i.e., some topological inherent directly from `Topology` and others inherent from the `MultihopTopo`intermediate class creates confusion.

We should have a consistent abstraction for all topology types while maintaining backwards compatibility. This design pattern does not meaningfully improve the topology architecture.

PR #300, opened 2026-02-11
Updating the Distributed Routing Protocol, API Doc, and Tutorial

caitaozhan:
1. Fix some logical bugs in the DistributedRoutingProtocol
   1.  When a node install a newer LSA into the LSDB, the node needs to check and satisfy the pending_requests across all neighbor FSMs -- because LSDB is global while the FSM is per neighbor and when LSDB is updated, all FSMs needs to get checked
   2.  DBD packet retransmission is added because sometimes a node will ingore responding to DBD packets when the FSM state is {Down, Init, TwoWay}
   3.  ACK is changed to based on receipt, not based on whether the LSU packet let to a LSDB change. A new unit test is added. Some minor refactor in timeline.py

2. Updated the API doc. Now the remote Read the Docs will always stay updated. No need for manually updating the Read the Docs. The `source/references` are completely removed

3. Update the Chapter 4 of the Resource Manager and the Chapter 5 of the tutorial about the network manager

### DIFF
--- a/docs/scripts/generate_reference_rst.py
+++ b/docs/scripts/generate_reference_rst.py
@@ -110,7 +110,16 @@ def discover_modules(source_dir: Path, relative_source_dir: str) -> list[str]:
 def write_module_rst(module_path: str, output_file: Path) -> None:
     module_name = module_path.split(".")[-1]
     title = get_title(module_name)
-    content = f"{title}\n{'=' * len(title)}\n\n.. automodule:: {module_path}\n    :members:\n"
+    extra_options = ""
+    if module_path == "sequence.constants":
+        extra_options = "    :undoc-members:\n"
+
+    content = (
+        f"{title}\n{'=' * len(title)}\n\n"
+        f".. automodule:: {module_path}\n"
+        f"    :members:\n"
+        f"{extra_options}"
+    )
     output_file.write_text(content, encoding="utf-8")
 
 
--- a/sequence/network_management/routing_distributed.py
+++ b/sequence/network_management/routing_distributed.py
@@ -8,7 +8,7 @@ from collections import defaultdict
 from dataclasses import dataclass, field
 from enum import Enum, auto
 from heapq import heappop, heappush
-from typing import TYPE_CHECKING, Union
+from typing import TYPE_CHECKING, Iterator
 
 if TYPE_CHECKING:
     from sequence.topology.node import QuantumRouter
@@ -19,8 +19,6 @@ from ..kernel.process import Process
 from ..message import Message
 from ..protocol import Protocol
 from ..utils import log
-
-MAX_AGE = 1000 * SECOND  # maximum age of LSA in seconds
 
 
 class DistRoutingMsgType(Enum):
@@ -217,7 +215,7 @@ class LinkStateDB:
     def get(self, adv: str) -> LSA | None:
         return self.lsas.get(adv, None)
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator[LSA]:
         return iter(self.lsas.values())
 
     def install(self, lsa: LSA, now: int) -> bool:
--- a/.gitignore
+++ b/.gitignore
@@ -148,3 +148,6 @@ tmp/
 utils/json_config_generators/*.json
 
 .virtual_documents/
+
+# docstring files
+*.rst
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.8.5] - 2026-2-11
+## [0.8.5] - 2026-2-16
 ### Added
-- Class `DistributedRoutingProtocol` is added, among many other supporting classes in `routing_distributed.py`. This module do truly distributed entanglement routing.
+- Class `DistributedRoutingProtocol` is added, among many other supporting classes in `routing_distributed.py`. This module performs truly distributed entanglement routing.
 - Class `NetworkManager` has two new attributes: `forwarding_table` and `routing_protocol`.
 - Add cutoff flag to memory to allow disable of expiration
 - Add a template in config json for network manager to select routing protocols (distributed vs. static)
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -34,19 +34,6 @@ Welcome to the SeQUeNCe documentation page
     references/resource_management/top
     references/topology/top
 
-.. toctree::
-    :maxdepth: 1
-    :caption: Parallel Simulation Pages:
-
-    parallel/install/prerequisite
-    parallel/explanation/parallel_tl_qm
-    parallel/guide/running_script
-
-.. toctree::
-    :maxdepth: 1
-    :caption: Parallel Module References:
-
-    references/parallel/top
 
 Indices and tables
 ==================
--- a/sequence/constants.py
+++ b/sequence/constants.py
@@ -1,36 +1,58 @@
-"""useful constants"""
+"""Shared constants used across SeQUeNCe.
+
+This module centralizes immutable values that are reused by simulation and
+network components, including:
+
+- Physical constants (for example, speed of light in meters per picosecond).
+- Canonical quantum state vectors (basis states and Bell states).
+- Numerical tolerances (`EPSILON`) for floating-point comparisons.
+- Time unit conversion factors in picoseconds (`NANOSECOND` to `SECOND`).
+- Built-in formalism and protocol identifier strings.
+"""
 
 from typing import Final
 
-# speed of light in (m / pico second)
+#: Speed of light in meters per picosecond.
 SPEED_OF_LIGHT: Final = 2e-4
 
-# |0> and |1>
+#: Qubit computational basis state |0>.
 KET0: Final = (1, 0)
+#: Qubit computational basis state |1>.
 KET1: Final = (0, 1)
 
-# four Bell states
+#: Normalization factor 1/sqrt(2) used by Bell states.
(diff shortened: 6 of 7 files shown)


sequence/network_management/routing_distributed.py
```diff
@@ -50,88 +51,99 @@ class LSAHeader:
...
+
     sender: str
-    acks: list[tuple[str, int]]  # list of (advertising_router, seq_number)
+    acks: list[
+        tuple[str, int]
+    ]  # list of (advertising_router, seq_number) that are updated
 
 
-DistRoutingPayload = Union[HelloPayload, DBDPayload, LSRPayload, LSUPayload, LSAckPayload]
+DistRoutingPayload = Union[
```
hayekr: No need to use Union anymore, use "HelloPayload | DBDPayload..."

sequence/network_management/routing_distributed.py
```diff
@@ -244,20 +296,22 @@ def purge_withdrawn(self, now: int) -> list[str]:
...
-        fsm (dict[str, NeighborFSM]): mapping of neighbor name to its FSM.
+        fsm (dict[str, NeighborFSM]): mapping of all neighbors' name to its FSM.
         link_cost (dict[str, float]): mapping of neighbor name to link cost.
         adj_cost (dict[str, float]): mapping of neighbor name with 2-way hellos to link cost.
         seq_number (int): sequence number for own LSA.
         refresh_enabled (bool): whether refreshing own LSA is enabled.
         last_originated_time (int): time of last originated LSA.
     """
+
     HELLO_INTERVAL = 1 * SECOND  # interval between HELLOs
```
hayekr: All of these should by typeset to ensure they are ints.

sequence/network_management/routing_distributed.py
```diff
@@ -1,47 +1,48 @@
...
 
 class DistRoutingMsgType(Enum):
     """Enum class for message types used in distributed routing protocol."""
-    HELLO = auto() # HELLO message
-    DBD = auto()   # DATA BASE DESCRIPTION
-    LSR = auto()   # LINK STATE REQUEST
-    LSU = auto()   # LINK STATE UPDATE
-    LSAck = auto() # LINK STATE ACKNOWLEDGEMENT
+
+    HELLO = auto()  # HELLO message
```
caitaozhan, line 27: I prefer vertical alignment. It looks more beautiful

sequence/network_management/routing_distributed.py
```diff
@@ -50,88 +51,99 @@ class LSAHeader:
...
-    
+
     Attributes:
         sender (str): name of sender node.
         acks (list[tuple[str, int]]): list of (advertising_router, seq_number) acknowledgements.
     """
+
     sender: str
-    acks: list[tuple[str, int]]  # list of (advertising_router, seq_number) that are updated
+    acks: list[
```
caitaozhan: Why break one line into three lines? It looks terrible.

sequence/network_management/routing_distributed.py
```diff
@@ -234,26 +260,34 @@ def install(self, lsa: LSA, now: int) -> bool:
             return True
         else:
             # install if newer
-            if lsa.header.seq_number > existing_lsa.header.seq_number:    # larger seq number -> newer LSA
+            if (
```
caitaozhan: why break one line into 3 lines? it looks soo ugly

sequence/network_management/routing_distributed.py
```diff
@@ -234,26 +260,34 @@ def install(self, lsa: LSA, now: int) -> bool:
...
+
     def purge_withdrawn(self, now: int) -> list[str]:
-        """Purge withdrawn LSAs from database.
+        """Purge withdrawn LSAs from the database.
 
         Returns:
             list[str]: list of advertising routers whose LSAs are purged.
         """
-        to_purge = [adv for adv, lsa in self.lsas.items() if now - lsa.header.originated_time >= DistributedRoutingProtocol.MAX_AGE]
+        to_purge = [
```
caitaozhan: This looks insane

sequence/network_management/routing_distributed.py
hayekr: It was the auto linter. I can go back and fix this

sequence/network_management/routing_distributed.py
hayekr: it shouldn't have done that

sequence/network_management/routing_distributed.py
hayekr: I agree

sequence/network_management/routing_distributed.py
hayekr: I agree

docs/scripts/generate_reference_rst.py
```diff
@@ -0,0 +1,174 @@
...
+def prune_stale_files(output_dir: Path, expected_files: set[str]) -> None:
+    for rst in output_dir.glob("*.rst"):
+        if rst.name not in expected_files:
+            rst.unlink()
+
+
+def main() -> None:
+    repo_root = Path(__file__).resolve().parents[2]
+    sequence_root = repo_root / "sequence"
+    refs_root = repo_root / "docs" / "source" / "references"
```
hayekr, line 142: Assume it doesnt exist and create it. refs_root.mkdir(exists_ok=True)

caitaozhan:
The `latest` is currently tracking this branch: https://sequence-rtd-tutorial.readthedocs.io/latest/

docs/scripts/generate_reference_rst.py
```diff
@@ -0,0 +1,156 @@
...
+        title="Entanglement Management",
+        description="The Entanglement Management module provides protocols for generating, purifying, and swapping quantum entanglement.",
+    ),
+    SectionSpec(
+        output_dir="kernel",
+        source_dir="kernel",
+        title="Kernel",
+        description="The Kernel module provides the discrete event simulator for SeQUeNCe.",
+    ),
+    SectionSpec(
```
hayekr, line 44: How does this misc work. There is no directory called misc/

docs/scripts/generate_reference_rst.py
```diff
@@ -0,0 +1,156 @@
...
+        output_dir="misc",
+        source_dir=".",
+        title="Miscellaneous",
+        description="Miscellaneous SeQUeNCe modules.",
+        static_entries=("utils/top",),
```
hayekr, line 49: Where is utils/top?

sequence/network_management/routing_distributed.py
```diff
@@ -285,27 +318,27 @@ def init(self):
         if self.refresh_enabled:
             process = Process(self, "refresh_lsa", [self.MAX_AGE // 2])
             time = self.owner.timeline.now() + self.MAX_AGE // 2
-            event = Event(time, process)
+            event = Event(time, process, priority=self.owner.timeline.schedule_counter)
```
hayekr, line 319: Why is the priority set to the schedule counter

sequence/network_management/routing_distributed.py
```diff
@@ -181,17 +210,18 @@ class LinkStateDB:
...
     def __init__(self):
         self.lsas: dict[str, LSA] = {}
-    
+
     def get(self, adv: str) -> LSA | None:
         return self.lsas.get(adv, None)
-    
-    def __iter__(self) -> iter:
+
+    def __iter__(self):
```
hayekr: Consider returning the type 

def __iter__(self) -> Iterator['LSA']:

sequence/network_management/routing_distributed.py
```diff
@@ -244,20 +275,22 @@ def purge_withdrawn(self, now: int) -> list[str]:
...
         adj_cost (dict[str, float]): mapping of neighbor name with 2-way hellos to link cost.
         seq_number (int): sequence number for own LSA.
         refresh_enabled (bool): whether refreshing own LSA is enabled.
         last_originated_time (int): time of last originated LSA.
     """
-    HELLO_INTERVAL = 1 * SECOND  # interval between HELLOs
-    DEAD_INTERVAL  = 4 * SECOND  # time to declare neighbor dead
-    MAX_AGE = 1000 * SECOND      # maximum age of LSA
+
+    HELLO_INTERVAL: int = 1 * SECOND  # interval between HELLOs
```
hayekr, line 288: Should these constants be class attributes? globals?

sequence/network_management/routing_distributed.py
caitaozhan: The priority is infinity if no value is passed in. So if priority is not set, what could happen is multiple events with the same time stamp all have the same priority (infinity). This could lead to unwanted behavior.

schedule_counter is a counter that increases by 1 each time an event is scheduled. Setting it to the schedule_counter only helps (serve as break even condition when the time stamps are the same), won't hurt.

sequence/network_management/routing_distributed.py
caitaozhan: I view these attributes as "protocol parameters" and part of the `DistributedRouting` class behavior. So, I put them inside the class.

hayekr: That's alright but why are they not class attributes, I.e., self.HELLO_INTERVAL

docs/scripts/generate_reference_rst.py
caitaozhan: Here it is: https://sequence-rtd-tutorial.readthedocs.io/stable/references/misc/top.html

It is a collection of `constants`, `message`, `protocol`, and `util`

docs/scripts/generate_reference_rst.py
caitaozhan: Here it is: https://sequence-rtd-tutorial.readthedocs.io/latest/references/misc/utils/top.html

sequence/network_management/routing_distributed.py
caitaozhan: HELLO_INTERVAL and three others are class attributes because they are defined outside the `__init__()`

PR #301, opened 2026-02-11
Migrate rule architecture to ResourceManager

hayekr:
This PR mainly performs a refactor that decouples the RSVP ResourceReservationProtocol from the ResourceManager.

Main Function: Network Manager and Resource Manager now subscribe to the following contract. The network manager schedules future resource usage on its timecard. This timecard is like a calendar which contains information of when there has been an approved reservation. This timecard and reservation information is then relayed to the resource manager where it generates and loads the rules based on this timecard and reservation. This creates a clear split of responsibility: NetworkManager for reservation scheduling and relay of information, ResourceManager for storing actual MemoryInformation and execution of resource actions.

The MemoryTimecard will stay as this is the 'calendar', and is just used to inform the resource manager of the reservation details.

Summary of Changes.
- All actions, conditions, and request functions are moved to a new file: action_condition_set.py
- The ResourceReservationProtocol interfaces with the resource manager by passing its memory array name, timecard, path, and reservation.
- Moves rule creation from reservation to the resource manager (generate_load_rules). This is a two in one function which creates the rules and schedules them on the timeline.
- Rule creation, loading, and management is now solely the responsibility of the network manager.
-  The test suite for ResourceManager, RuleManager, Reservation is significantly extended.


docs/source/tutorial/chapter5/network_manager.md
JoaquinChung, line 1: Should we provide a rationale for why are we commenting out these lines, or should we just remove them?

JoaquinChung (commented):
This looks good to me.

docs/source/tutorial/chapter5/network_manager.md
hayekr: We will add it back in a later PR, just commented out while we decide what to do with it

JoaquinChung (dismissed):

sequence/network_management/reservation.py
```diff
@@ -437,8 +150,8 @@ def pop(self, src: str, msg: "ResourceReservationMessage"):
             path = [qcap.node for qcap in msg.qcaps]
             if self.schedule(msg.reservation):  # schedule success
                 if self.owner.name == msg.reservation.responder:
-                    rules = self.create_rules(path, reservation=msg.reservation)
-                    self.load_rules(rules, msg.reservation)
+                    self.owner.resource_manager.generate_load_rules(path, msg.reservation, self.timecards, self.memory_array_name)
+                    self.accepted_reservations.append(msg.reservation)
                     msg.reservation.set_path(path)
```
caitaozhan, line 155: Something to consider: 
Given that the `path` is an attribute of `reservation`, the `generate_load_rules()` does not need the argument `path`. We just need to `msg.reservation.set_path(path)` before calling the `generate_load_rules()`

For `generate_load_rules()`, 4 arguments are down to 3. Not a big deal, though.

hayekr: This is untrue for APPROVE RSVP messages, where the msg.path is not part of the reservation. Reducing arguments from 4 to 3 would require introducing additional logic. Alternatively, we could use set_path at approval, changing the reservation path.

hayekr: Resolving, merging no change. Will address at later date

hayekr:
> Some aesthetic comments for `action_condition_set.py`.
> 
>     * Some function signatures are split into 3 lines of code, some are not. The function signatures don't appear long enough to split. Is there a threshold for the line length?
> 
>     * The big if statements. The big if statements were hard to read before; now they are hard to read + two extra lines due to your PyCharm auto. I'm thinking of wrapping up the conditions in the big if statements into a nested helper function and giving the helper function a meaningful name.
> 
>     * We shall add the docstrings (i.e., description of the Args) for the functions.

Yeah we should look into this later. These if-statements go beyond the length limit of a line; we should find a way to refactor this file.

sequence/network_management/reservation.py
caitaozhan: Fix: the msg.path **is** part of the reservation. Because reservation.path is already set at the responder. The same message passes from the responder back to the initiator.

But anyway, reducing 4 arguments to 3 is not a big deal

caitaozhan (approved):
Looks good. Let's merge!

PR #303, opened 2026-02-12
Update the readme.md

hayekr:
This pull request updates the installation instructions in the `README.md` to use `uv sync` instead of `pip install --editable`, and adds a new citation to the list of related papers.

Documentation updates:

* Updated the installation instructions to use `uv sync` instead of `pip install --editable . --config-settings editable_mode=strict`, and removed the note about `make install_editable` for Linux and Mac users.

Citation additions:

* Added a new citation for H. Miller et al., "Simulation of a Heterogeneous Quantum Network" (2025) to the list of related papers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,20 +36,46 @@ SeQUeNCe is an open source, discrete-event simulator for quantum networks. As de
 
 These modules can be edited by users to define additional functionality and test protocol schemes, or may be used as-is to test network parameters and topologies.
 
-## Installing
-SeQUeNCe requires [Python](https://www.python.org/downloads/) 3.11 or later. You can simply install SeQUeNCe using `pip`:
+## Installation
+### For Users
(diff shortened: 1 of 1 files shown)


JoaquinChung (commented):
I agree that for legacy users, we should keep the `pip install --editable` version. What I'm debating is whether we should provide all the examples of setting the venv or just say that the pip option is available?

hayekr:
> I agree that for legacy users, we should keep the `pip install --editable` version. What I'm debating is whether we should provide all the examples of setting the venv or just say that the pip option is available?

Developers who aim to contribute to the repository are required to keep their repositories up to date. Therefore, they should use the build-backend we provide. In development there shouldn't be any legacy users, developers should keep their branches tracking RnD or master. However, end-users may want to stay legacy, and that is where they can do `pip install sequence`.

PR #304, opened 2026-02-25
[Minor] Moving things around and removing "\"

caitaozhan:
1. Move function `routing_protocol.init()` from `RouterNetTopo._generate_forwarding_table()` to `QuantumRouter.init()`
2. Move attribute `protocol_type` from `EntanglementProtocol` to `Protocol`, which is should be the right place. For example, the `DistributedRoutingProtocol` is a subclass of `Protocol` and also need a `protocol_type`
3. Replace the `\` for new lines with `()`, etc.

### DIFF
--- a/sequence/entanglement_management/generation/generation_base.py
+++ b/sequence/entanglement_management/generation/generation_base.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from math import sqrt
-from typing import TYPE_CHECKING, List, Dict, Type, Any
+from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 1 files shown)


sequence/topology/node.py
```diff
@@ -417,6 +417,7 @@ def init(self):
         Inherit parent function.
         """
         super().init()
```
hayekr: This function is pass in `Node` class, this does nothing,

caitaozhan: Shall this line be removed?

hayekr: yes

Issue #241, opened 2025-08-26
Implement registry pattern for user-defined Node children without source modifications


hayekr:
Currently, the preferred way to implement custom protocols for experiments is to create new subclasses of existing internal objects and modify them as needed. However, this approach requires direct modifications to the source code, which can lead to increased maintenance overhead and coupling between user extensions and core components.

**Motivation:**
For example, the `RequestApp` creates a `QuantumRouter(Node)` with rules generated by `ResourceReservationProtocol.create_rules`. To change the rules, users must define a new Node subclass and manually modify the source code to apply new rules, which often requires updating multiple dependencies and files.

**Proposed Solution:**
Introduce a registry pattern, enhanced with decorator functionality, to enable user-side registration of new Node children. This would allow users to extend or register new node types and associated behaviors without modifying the core codebase. The registry should seamlessly integrate user-defined nodes with the original class and protocol logic, significantly easing extensibility for experimental protocols.

**Benefits:**
- Reduces the need for direct source code changes by users.
- Simplifies the process of adding experimental or custom protocol logic.
- Promotes a more modular and maintainable architecture for protocol development.

This feature will make it much easier for users to implement and test new experimental protocols with custom nodes, lowering the barrier for contribution and experimentation.

This will be identical in function to what issue 231 accomplishes.

PR #313, opened 2026-03-03
Topology  Refactor (UX improvements pt. 1)

asmuley2:
## Summary

Full refactor of `sequence/topology` to consolidate duplicated build logic, improve extensibility, and modernize config UX.

### The build pipeline problem

Every concrete topology (`RouterNetTopo`, `DQCNetTopo`, `QlanStarTopo`) was duplicating nearly identical build pipelines. `Topology` now owns the shared pipeline as a template. Family-specific behavior is composed through `TopologyFamily` implementors — an aggregation-based bridge pattern where the concrete topology passes back a family object to its parent, and that family provides optional hooks to influence each stage of the build.

### Why this matters for extensibility

The relationship between `Topology` and `TopologyFamily` follows an aggregation-based bridge pattern — the abstraction side (`Topology` and its subclasses) delegates family-specific behavior to a composed implementor (`TopologyFamily` and its subclasses), but neither side owns the other's lifecycle. This means the two hierarchies can evolve independently.

Because `TopologyFamily` subclasses only override the hooks they care about, new families can be built by minimally extending an existing one. If the behavior difference is small (e.g., a different midpoint selection strategy), you subclass the closest family. If it's fundamentally different, you subclass `TopologyFamily` directly.

I have a UML diagram showing this structure that I'll present at Friday's meeting.

### Config UX improvements

- Dict-based construction alongside file-based loading
- `**kwargs` override/extend file configs at construction time (scalars replaced, lists appended)
- YAML support alongside JSON
- Centralized constants in `const_topo.py`. Old access paths (e.g., `RouterNetTopo.QUANTUM_ROUTER`) still work but emit `DeprecationWarning` pointing to the new import. This is implemented via a `_DeprecatedAttrMeta` metaclass on the concrete topology classes. It works but is noisy — 16 warnings in the current test suite. Open to changing the deprecation approach if there's a preferred pattern.
- Caller-owned configs are never mutated internally

### What this PR sets up for next

Node addition currently uses if/elif dispatch keyed on centralized type constants from `const_topo.py`. Node classification (e.g., "is this a midpoint?" or "is this a routing endpoint?") is handled by direct string comparison against these same constants. Both patterns are functional but not user-extendable — adding a custom node type still means editing core if/elif chains and classification checks. A follow-up PR will replace both with a registry pattern (issue #241), where nodes self-declare their roles and the pipeline dispatches through the registry rather than string matching.

## Other fixes and improvements

- Fixed `DQCNetTopo.infer_memory_owners` using node type strings (`"DQCNode"`) instead of actual node names — would `KeyError` at runtime
- Added missing `memory_expire` callback on `QlanClientNode` and `QlanOrchestratorNode` — without it, memory expiration would `AttributeError`
- Fixed `QlanClientNode` error message saying "minimum" when it meant "maximum"
- Removed stray debug `print()` calls from `BSMNode`, `QKDNode`, and `QlanOrchestratorNode`; moved diagnostic info into exception messages
- Explicit error if a BSM midpoint doesn't connect to exactly 2 endpoints (was a cryptic unpack error)
- Validated `cconnections` exist for all `qconnections` before midpoint expansion
- `get_timeline()` uses `RuntimeError` instead of `assert` (asserts are stripped under `python -O`)
- Unified `stop_time` default to `float('inf')` across all topologies (was `10**23` in `RouterNetTopo`)
- Node classification (midpoint vs endpoint) via centralized type constants in `const_topo.py`
- Fixed QLAN's double-pass node construction by introducing an `_order_nodes` hook on `TopologyFamily` (this was necessary to streamline the build into a single pass over the node list; any new family can now define its own build order, which provides significant extensibility)
- Broke qconnection expansion into smaller hooks (`_midpoint_node_config`, `_qchannel_configs_for_qconnection`, `_cchannel_configs_for_qconnection`)
- Unified QLAN memory handling around `MemoryArray` with template-driven config (legacy flat-key format still supported). Without this unification, every node family needs its own memory handling path, leading to subclass explosion. As a companion fix, `Node.get_components_by_type("Memory")` now also searches inside `MemoryArray` containers so callers still find individual memories.
- Cleaned up DQC helper logic for correct ownership/mapping source of truth

## Testing

Added or updated tests for YAML/JSON parity, dict-based construction, kwargs override/extension, config-loading edge cases, and new sample topology configs.


hayekr:
Hi @asmuley2, thank you for your contribution. To ease the review process can you please place the second issue's changes into a separate PR. A PR should ideally cover one single issue.
Thank you.

asmuley2:
Hi @hayekr, thank you for the quick feedback. The topology refactor can stand on its own, but the node registry PR would build directly on top of it, meaning the two can't be made in parallel; they'd have to be sequential. I'd appreciate some guidance on how to break it up.

Should I reuse this PR for the topology refactor and trim the node registry changes out, or close this one and open two new PRs? If two new PRs, should I open the node registry PR after the refactor PR gets approved, or shortly after submitting it? Either way, I'd add comments noting the parts that the node registry PR will improve on (scoped to the topology module, no changes to the node modules or their subclasses).

Additionally, I have some examples demonstrating the benefits and use cases of these improvements. Should I add those as another commit on top of the two I mentioned, or would it be more appropriate to include them as a separate PR? Or should I just present them at the next meeting and move on?

Regardless, I'll have a lot to present and announce at the next meeting.

AM

asmuley2:
After thinking about it, here's my plan:

I'll reuse this PR for the topology refactor only. I'll trim out the node registry changes and include comments on the parts of this refactor that will be streamlined by the node registry pattern PR. I'll also update the description accordingly. Then I'll open a separate PR for the node registry work (issue #241) on top of the refactor.

Please let me know if this works.

AM

hayekr:
Sounds good, I would cherry pick the relevant changes to issue #241 to another branch and submit Pr for those changes. Ideally you submit a PR for refactor topology, and an additional PR for additions you intend to make to Node, etc.

asmuley2:
This PR has been refocused on the topology refactor. A companion PR will follow with registry-based changes to Node.

sequence/topology/topology.py
```diff
@@ -4,121 +4,313 @@
...
+        if kwargs:
+            self._merge_overrides(config, kwargs)
+        if ALL_NODE not in config or not config[ALL_NODE]:
+            raise ValueError("Config must contain a non-empty 'nodes' list.")
+        self._setup(config, family)
+        self._input_cfg = copy.deepcopy(source_config)
+        self._build_cfg = config
+
+    @staticmethod
+    def _load_config(path: str) -> dict:
```
hayekr: Python really doesnt have the concept of private. Remove all `_`.

hayekr: It really should be used sparingly, to signal that something may be pseudo-private

sequence/topology/topology.py
```diff
@@ -4,121 +4,313 @@
...
+            raise TypeError(
+                f"config must be a file path (str) or a config dict, got {type(config).__name__}"
+            )
+        config = copy.deepcopy(source_config)
+        if kwargs:
+            self._merge_overrides(config, kwargs)
+        if ALL_NODE not in config or not config[ALL_NODE]:
+            raise ValueError("Config must contain a non-empty 'nodes' list.")
+        self._setup(config, family)
+        self._input_cfg = copy.deepcopy(source_config)
```
hayekr: `self._input_csg` doesnt need pseudo-private prefix, and is never mutated; thus deepcopy is unnecessary. Additionally, `source_config` doesn't appear to be mutated either, further negating the need for a deep-copy.

sequence/topology/topology.py
```diff
@@ -4,121 +4,313 @@
...
+        """
+        if isinstance(config, str):
+            source_config = self._load_config(config)
+        elif isinstance(config, dict):
+            source_config = config
+        else:
+            raise TypeError(
+                f"config must be a file path (str) or a config dict, got {type(config).__name__}"
+            )
+        config = copy.deepcopy(source_config)
```
hayekr: Setting to this seems unnecessary. `source_config` gets the passed `config`; then we deepcopy `source_config` back to `config`; then mutate in-place `config`. `self._input_cfg` is never used again except once as an iterator in dqc. 

I suggest: save raw_config and merged_overrides_coinig as two separate config objects; dont worry about doing this inplace. `merge_overrides` should simply take in a config and kwargs, and return the new config, which can be set to a new variable. No need for deepcopy or inplace modification.

sequence/topology/topology.py
```diff
@@ -4,121 +4,313 @@
...
+        if kwargs:
+            self._merge_overrides(config, kwargs)
+        if ALL_NODE not in config or not config[ALL_NODE]:
+            raise ValueError("Config must contain a non-empty 'nodes' list.")
+        self._setup(config, family)
+        self._input_cfg = copy.deepcopy(source_config)
+        self._build_cfg = config
```
hayekr: Same logic here, its the same data repeated with wasteful deepcopy.

sequence/topology/topology.py
```diff
@@ -4,121 +4,313 @@
...
+
+    @staticmethod
+    def _load_config(path: str) -> dict:
+        """Load a JSON or YAML config file and return the parsed dict."""
+        if not path.endswith(('.json', '.yaml', '.yml')):
+            raise ValueError(
+                f"Unsupported config file format: {path}. "
+                "Use .json or .yaml/.yml"
+            )
+        with open(path) as fh:
```
hayekr: wrap this in try and handle possible exceptions.

sequence/topology/topology.py
```diff
@@ -4,121 +4,313 @@
...
+            if path.endswith(('.yaml', '.yml')):
+                import yaml
```
hayekr: Imports need to be at the top

sequence/topology/topology.py
```diff
@@ -4,121 +4,313 @@
...
+    def _merge_overrides(build_config: dict, extra_config: dict) -> None:
+        """Merge extra config values into a build config dict in place.
+
+        Lists are extended, dicts are updated, everything else is overridden.
         """
+        _LIST_KEYS = {
+            'nodes': ALL_NODE, 'qconnections': ALL_Q_CONNECT,
+            'cconnections': ALL_C_CONNECT, 'qchannels': ALL_Q_CHANNEL,
+            'cchannels': ALL_C_CHANNEL,
+        }
```
hayekr, line 154: Just have a `merged_config = {}` and build it instead of mutating `build_config` in-place. Doing this we can remove deepcopy

sequence/topology/topology.py
```diff
@@ -4,121 +4,313 @@
...
+        for friendly, const in _LIST_KEYS.items():
```
hayekr: This is unclear var names. Just stick to `key, val`.

sequence/topology/topology.py
```diff
@@ -4,121 +4,313 @@
...
+                return yaml.safe_load(fh)
+            return json.load(fh)
+
+    @staticmethod
+    def _merge_overrides(build_config: dict, extra_config: dict) -> None:
+        """Merge extra config values into a build config dict in place.
+
+        Lists are extended, dicts are updated, everything else is overridden.
         """
+        _LIST_KEYS = {
```
hayekr: This doesnt need pre-fixed underscore

sequence/topology/topology.py
```diff
@@ -4,121 +4,313 @@
...
+                distance,
+                delay,
+            )
+            cc_obj.set_ends(src_obj, dst_str)
+            self.cchannels.append(cc_obj)
 
     def get_timeline(self) -> "Timeline":
-        assert self.tl is not None, "timeline is not set properly."
+        if self.tl is None:
+            raise RuntimeError("Timeline is not set — topology may not be fully initialised.")
```
hayekr: initialized is spelled wrong.

hayekr (changes requested):
I will continue reviewing once tests pass. Please ensure pytest passes prior to submitting a PR. Please review comments made thus far.

asmuley2:
@hayekr Thank you for your feedback. I apologize for the errant code in the current commit.

Interestingly, the Part 2 PR passed all tests on the first attempt, which suggests an issue may have occurred during the revert to the pre-Node registry pattern state. If you re-running the test suite against my latest commit all the tests should pass. 

In the meantime, I'll work on your suggestions.

asmuley2:
I have converted the pt. 2 PR to a draft and addressed all the comments thus far. I can respond to them individually once I get the chance

sequence/topology/topology.py
asmuley2: I removed most of them with the exception of some pseudo-private functions that emerged from the DRY part of this refactor (i.e. in Topology, _validate_qconnection_dependencies, _calc_cc_delay, and _make_classical_channel). I also chose to preserve preexisting _ conventions, like _add_qchannels, _add_qconnections, _add_cchannels, and _add_cconnections. Let me know if you wanted something different.

sequence/topology/topology.py
asmuley2: I agree that the deep copy is unnecessary. The reason I did it was because DQCNetTopo had a method that required the config dict to only contain DQCNodes, which isn't the case after bsm midpoints are automatically added when the topology is initialized. I decided to fix this issue on the DQC side to skip over anything that isn't a DQCNode. This is correct, since every subclass of Topology should be using the config only after it is fully expanded.

The only time I make a deepcopy of a dict is when we pass one in as the base config. This way the Topology's ownership of its config dict is identical to the case that we load a config file.

sequence/topology/topology.py
asmuley2: Fixed

sequence/topology/topology.py
asmuley2: Fixed

sequence/topology/topology.py
asmuley2: no longer an issue - removed unnecessary deepcopies.

sequence/topology/topology.py
asmuley2: fixed

sequence/topology/topology.py
asmuley2: fixed

sequence/topology/topology.py
asmuley2: fixed

hayekr:
Closing PR, will not merge.

PR #314, opened 2026-03-04
 Node Registry (UX improvements pt. 2)

asmuley2:
## Summary

Adds a node registry pattern to decouple node addition from the topology build pipeline (issue #241). Builds on the topology refactor PR #313.

### The problem

After the topology refactor, node addition uses if/elif chains and node classification uses string comparisons against centralized constants. This works for built-in types but isn't user-extendable — adding a custom node type means editing core dispatch and classification logic.

### The solution

The registry pattern replaces both: nodes self-register via `@Node.register("TypeName")` and declare their topology roles (e.g., `topology_roles = {ROLE_BSM_ENDPOINT, ROLE_ROUTABLE_ENDPOINT}`), so the pipeline dispatches and classifies through the registry rather than hardcoded checks. Each registered node implements a `from_config(...)` classmethod that the build pipeline calls via `Node.create(type_string, ...)`. No fallthrough cases, no core edits needed.

In practice this is simple. From `example/topology_extensions/custom_router_node.py`:

```python
@Node.register("YbRouterNode")
class YbRouterNode(QuantumRouter):
    platform = "Yb"
    operating_wavelength = 980
```

That type name works directly in config dicts or JSON files — the pipeline builds it like any built-in node. For nodes that need custom construction logic, implement `from_config`. From `example/topology_extensions/registry_pipeline_demo.py`:

```python
@Node.register("InstrumentedRouter")
class InstrumentedRouter(QuantumRouter):
    def __init__(self, name, tl, memo_size=50, seed=None,
                 component_templates=None, gate_fid=1, meas_fid=1):
        super().__init__(name, tl, memo_size, seed, component_templates, gate_fid, meas_fid)
        self.eg_attempt_count = 0

    @classmethod
    def from_config(cls, name, tl, config, template, **kwargs):
        return cls(name, tl, config.get("memo_size", 50), config.get("seed"),
                   component_templates=template)
```

## Examples

Added `example/topology_extensions/` with three demos:
- **`custom_router_node.py`** — simplest path: register a node subclass and use it in a dict config
- **`registry_pipeline_demo.py`** — file-based config with kwargs overrides and a custom `from_config`
- **`heterogeneous_midpoints.py`** — custom `TopologyFamily` subclass that auto-selects midpoint hardware based on endpoint wavelengths

## Testing

Added tests for node registry/factory behavior, including registration, dispatch, `from_config` integration, and error messages for unregistered types.


hayekr:
Closing PR, changes were not planned, will not merge.

Issue #310, opened 2026-03-02
Lack of support of BDS for Swapping


caitaozhan:
Add BDS for Swapping. Also, use the registry decorator.

PR #315, opened 2026-03-04
Entanglement Swapping BDS Implementation

caitaozhan:
https://github.com/sequence-toolbox/SeQUeNCe/issues/310. Created the `swapping` module that has the following classes. The subclasses use the registry decorator.

1. Base class: `EntanglementSwappingA` & `EntanglementSwappingB`
2. Subclass for ket vector and density matrix: `EntanglementSwappingA_Circuit` & `EntanglementSwappingB_Circuit`
3. Subclass for Bell diagonal state: `EntanglementSwappingA_BDS` & `EntanglementSwappingB_BDS`

Some other updates:
* Renamed the constant `KET_STATE_FORMALISM` to `KET_VECTOR_FORMALISM` in `sequence/constants.py` 
* Update docstrings, including clarifying the docstring for the `schedule_reservation` method in `request_app.py` to specify that it is called by both the initiator and responder when a reservation is approved.
* Some minor refactoring (i.e., removing the quotation marks for the typing)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0]
+### Added
+- Created the `swapping` module that has the following classes. The subclasses use the registry decorator (factory pattern).
+    - Base class: `EntanglementSwappingA` & `EntanglementSwappingB`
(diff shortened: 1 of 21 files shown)


hayekr:
Can you bump the version and update changelog

hayekr:
Also I think each sub_dir within entanglement_management needs __init__.py

hayekr (dismissed):
This PR introduces the an entanglement swapping protocol for the BDS formalism. It adds a factory pattern to manage the different swapping protocols. Minor changes include renaming of ket vector constant and the location of protocol_type.

hayekr (approved):

PR #322, opened 2026-03-10
Resource Manager Early Expire

caitaozhan:
A request has a `start_time` and `end_time`. When a request is finished before the `end_time`, then the reservation should expire early. If not, then the quantum network will keep generating entanglement pairs because the rules still exist.

Introducing EARLY_EXPIRE message type to fix it.


sequence/entanglement_management/swapping/swapping_bds.py
```diff
@@ -73,7 +73,7 @@ def start(self) -> None:
         assert self.right_memo.entangled_memory["node_id"] == self.right_node
 
         if self.owner.get_generator().random() < self.success_probability():
-            log.logger.debug(f'swapping successed!')
+            log.logger.info(f'swapping successed!')
```
hayekr: Can you fix the grammar. It should say "Swapping succeeded."

hayekr (changes requested):

hayekr (dismissed):

hayekr (approved):

hayekr (changes requested):
Tests are not passing

hayekr (approved):

Issue #325, opened 2026-03-12
Add a Routing Module Using Register Decorator


caitaozhan:
Create a parent class `RoutingProtocol` for the current classes `StaticRoutingProtocol` and `DistributedRoutingProtocol`.

Extensibility: the register decorator shall be used to make it easy to plug in new routing protocols in the future.

PR #326, opened 2026-03-13
Routing Module using Register

caitaozhan:
Added the `routing` module: a new parent class `RoutingProtocol` for the subclasses `StaticRoutingProtocol` and `DistributedRoutingProtocol`. The register decorator is used to make it easy to plug in new routing protocols in the future.

Issue https://github.com/sequence-toolbox/SeQUeNCe/issues/325


sequence/network_management/routing/routing_base.py
hayekr, line 1: Should there be a global type, or should we allow nodes to have differing routing protocols?

hayekr (dismissed):

sequence/network_management/routing/routing_base.py
caitaozhan: Good catch. The global_type is removed. I do not want to enforce that all nodes have the same routing protocol. Even though this might be the most common scenario.

hayekr (approved):

PR #332, opened 2026-04-10
Create new topology generation scripts

hayekr:
- Create module to convert arbitrary `nx.Graph` objects to sequence configuration objects.
- Port over all existing topologies to NetworkX; improves maintainability.
- `nx_converter` is an API which can be called by a script, and returns the sequence dict object, so JSON can be avoided.
- Create Fat Tree (k-ary, n-tree) topology
- Create BCube topology
- Create an all-in-one CLI with typer


caitaozhan (commented):
Looks good. But it would be helpful to give examples of how to use the command line. The examples could either be in the docstrings or in the utils/README.md

sequence/utils/nx_converter.py
```diff
@@ -0,0 +1,171 @@
...
+    if 'router_template' not in templates or 'bsm_template' not in templates:
+        raise ValueError("Template must contain 'router_template' and 'bsm_template' keys.")
+    output_dict: dict = {Topology.ALL_TEMPLATES: templates}
+
+    cc_delay_ps = cc_delay * 1e9
+    to_bsm_dist: float = qc_length * 1000 / 2  # Convert to meters, get middle
+
+
+    router_names = [router_name_func(i) for i in range(len(g.nodes))]
+    nodes: list[dict] = generate_nodes(router_names, 1, 'router_template',
```
caitaozhan: This argument is the memory size. It should be part of the input, instead of constant value 1.

sequence/utils/nx_converter.py
```diff
@@ -0,0 +1,171 @@
...
+        right_name = graph_to_name[right]
+        bsm_name = bsm_name_func(left_name, right_name)
+        bsm_nodes.append({Topology.NAME: bsm_name, Topology.TYPE: Topology.BSM_NODE, Topology.SEED: i, Topology.TEMPLATE: 'bsm_template'})
+
+        # Quantum Links (Node -> BSM <- Node)
+        qlinks.append({Topology.SRC: left_name, Topology.DST: bsm_name, Topology.DISTANCE: to_bsm_dist, Topology.ATTENUATION: qc_attn})
+        qlinks.append({Topology.SRC: right_name, Topology.DST: bsm_name, Topology.DISTANCE: to_bsm_dist, Topology.ATTENUATION: qc_attn})
+
+        # Classical Links (Node <-> BSM <-> Node)
+        clinks.append({Topology.SRC: left_name, Topology.DST: bsm_name, Topology.DISTANCE: to_bsm_dist, Topology.DELAY: cc_delay_ps})
```
caitaozhan, line 150: This classical link (node -> BSM) is never used in practice. Has anyone ever used it? If no, consider removing? I know it is not necessary, though.

sequence/utils/nx_converter.py
```diff
@@ -0,0 +1,171 @@
...
+        clinks.append({Topology.SRC: bsm_name, Topology.DST: left_name, Topology.DISTANCE: to_bsm_dist, Topology.DELAY: cc_delay_ps})
+        clinks.append({Topology.SRC: right_name, Topology.DST: bsm_name, Topology.DISTANCE: to_bsm_dist, Topology.DELAY: cc_delay_ps})
```
caitaozhan, line 152: This classical link (node -> BSM) is never used in practice. Has anyone ever used it? If no, consider removing? I know it is not necessary, though.

sequence/utils/config_generator_cli.py
```diff
@@ -0,0 +1,404 @@
...
+    formalism: Formalism = "bell_diagonal",
+    template_path: Template = "",
+    measurement_fidelity: MeasurementFidelity = 1,
+    gate_fidelity: GateFidelity = 1,
+) -> None:
+    template = get_template(template_path) if template_path else None
+    g = build_ring(nodes)
+    generate_config(
+        g,
+        qc_length=qc_length,
```
caitaozhan: There is a fundamental flaw for the waxman here. Because waxman gives you a random graph, the edge lengths cannot be a constant value -- the values vary. See how I deal with it: https://github.com/sequence-toolbox/SeQUeNCe/blob/64c218aeaae71518ae77fe06c49d39719c95cc39/utils/json_config_generators/config_generator_random.py#L105. 

Also, Waxman has its own parameters too, i.e., alpha, beta. See https://networkx.org/documentation/stable/reference/generated/networkx.generators.geometric.waxman_graph.html. See how I dealt with it: https://github.com/sequence-toolbox/SeQUeNCe/blob/64c218aeaae71518ae77fe06c49d39719c95cc39/utils/json_config_generators/config_generator_random.py#L44

It is hard to reuse the logic of others (line, mesh, etc) to waxman.

caitaozhan: To support graphs with varying edge lengths, the networkx graph object should include node coordinates. See how I deal with the random waxman: https://github.com/sequence-toolbox/SeQUeNCe/blob/64c218aeaae71518ae77fe06c49d39719c95cc39/utils/json_config_generators/config_generator_random.py#L31. It returns the V, E. The nodes in V have the (x,y) locations

caitaozhan:
It would be nice if the cli could take in a pickled nx.Graph object (the filename of the pickled nx.Graph object), where people create some arbitrary network topology that is not supported. The total number of topologies is infinite, whereas we can only support a limited number of common topologies.

For example, I need to create a topology in which edge lengths are not uniform. So for my nx.Graph object, I could add the weights to the edge, like `G.add_edge("A", "B", distance=2000)` and `G.add_edge("A", "B", distance=3000)`. Then, in the `nx_converter`, it can extract the weight of the edge and does not use the fixed uniform distance.

Also, `generate_classical(router_names, cc_delay)` is a joke because all classical links have the same delay. But there is no better way here. I always overwrite the classical delay on my own after the graph is initialized and apply some classical delay models.

sequence/utils/nx_converter.py
hayekr: We will keep, currently unused, use case is rare, but keeping causes negligible overhead

hayekr:
> It would be nice if the cli could take in a pickled nx.Graph object (the filename of the pickled nx.Graph object), where people create some arbitrary network topology that is not supported. The total number of topologies is infinite, whereas we can only support a limited number of common topologies.
> 
> For example, I need to create a topology in which edge lengths are not uniform. So for my nx.Graph object, I could add the weights to the edge, like `G.add_edge("A", "B", distance=2000)` and `G.add_edge("A", "B", distance=3000)`. Then, in the `nx_converter`, it can extract the weight of the edge and does not use the fixed uniform distance.
> 
> Also, `generate_classical(router_names, cc_delay)` is a joke because all classical links have the same delay. But there is no better way here. I always overwrite the classical delay on my own after the graph is initialized and apply some classical delay models.

This is resolved by allowing the CLI to take GML files as input. This is a standard format from NX

caitaozhan (approved):
Looks good! Useful tool

PR #336, opened 2026-05-06
Fix minor issues in Config CLI; Add sender_delay argument

caitaozhan:
1. Changed the `stop_time` argument default value in `config_generator_cli.py` to `None`. When the default value is `None`, then the stop time will be `10**23`, see https://github.com/sequence-toolbox/SeQUeNCe/blob/64c218aeaae71518ae77fe06c49d39719c95cc39/sequence/topology/router_net_topo.py#L58
2. Update the `cc_delay_ps` in `nx_converter.generate_config`
3. [major] update the delay model. Introducing the `sender_delay` at the sender side while sending the message, which includes, but is not limited to, processing delay, queueing delay, and transmission delay, etc.
4. Change the default delay behavior of the classical channel: when argument `delay <=0`, apply the `distance / speed_of_light`
5. Remove setting the node memory to the node degree because it contradicts the argument `memory_size`


hayekr (approved):
This PR fixes issues with delay calculation, removes the requirement that a node must have quantum memories at least the degree number, fixes CLI issues with default values.

This also introduces a sender delay to the classical channel, what amounts to processing delay.

PR #352, opened 2026-05-21
Fix single-qubit depolarizing noise channel

HasanTasdiq:
## Summary

`Noise.apply_depolarizing_noise` (in `sequence/utils/noise.py`) is documented to support 1 or 2 target qubits, but it hardcodes the Pauli index combinations as 2-qubit tuples:

```python
idx_tuples = [(i, j) for i in range(4) for j in range(4)]
```

For the single-qubit case (`k = 1`) only `idx_tuple[0]` is used, which produces two problems:

- **Spurious identity terms:** the tuples `(0, 1)`, `(0, 2)`, `(0, 3)` map the target qubit (and every other qubit) to `I`, so the full identity operator is added to the Pauli sum three times.
- **Wrong normalization:** the channel is divided by `len(pauli_ops) = 15` instead of `4^1 - 1 = 3`.

As a result the single-qubit depolarizing channel is mathematically incorrect. For example, applying `p = 0.3` to `|0><0|`:

| | output |
|---|---|
| current code | `diag(0.84, 0.16)` |
| correct channel `(1-p)ρ + (p/3)(XρX + YρY + ZρZ)` | `diag(0.8, 0.2)` |

## Fix

Generate the Pauli index tuples from the actual number of target qubits:

```python
idx_tuples = itertools.product(range(4), repeat=k)
```

- The **two-qubit** path is unchanged: 16 tuples, identity excluded → 15 operators (verified identical output to the previous implementation).
- The **single-qubit** path now yields the correct 3 operators (`X`, `Y`, `Z`) and matches the standard depolarizing channel.

## Test plan

- [x] Added `test_apply_depolarizing_noise_single_qubit` comparing the `k=1` output against the standard single-qubit depolarizing channel.
- [x] Existing `test_apply_depolarizing_noise` (two-qubit) still passes — the two-qubit path is unaffected.


caitaozhan (approved):
Looks good!

hayekr (approved):

PR #353, opened 2026-05-21
Fix POVM completeness for QSDetectorFockDirect

HasanTasdiq:
## Summary

In `QSDetectorFockDirect._generate_povms` (`sequence/components/detector.py`), the no-click POVM operator for **detector 1** is constructed from **detector 0's** click operator:

```python
povm1_1 = sum(series_elem_list)
povm1_0 = eye(truncation + 1) - povm0_1   # povm0_1 is detector 0's click operator
```

Because `povm1_0` is derived from `povm0_1` rather than `povm1_1`, detector 1's POVM pair `{povm1_0, povm1_1}` only sums to the identity when the two detectors share the same efficiency (so that `povm0_1 == povm1_1`). When the detectors have different efficiencies, `povm1_0 + povm1_1 != I`, so the measurement probability distribution used for photons arriving at input port 1 is incorrect.

The existing `test_QSDetectorFockDirect` sets both efficiencies to `1`, which masks the bug.

## Fix

Build detector 1's no-click operator from its own click operator:

```python
povm1_0 = eye(truncation + 1) - povm1_1
```

## Test plan

- [x] Added `test_QSDetectorFockDirect_povm_completeness`, which sets distinct detector efficiencies (0.9 and 0.5) and asserts that **both** detectors' POVM pairs sum to the identity.
- [x] Existing detector tests still pass.


caitaozhan (dismissed):
Thanks! Looks good.

caitaozhan (approved):
Looks good!

hayekr (approved):

PR #354, opened 2026-05-21
Fix density-matrix comparison in SingleAtomBSM (_eq_psi_plus)

HasanTasdiq:
## Summary

`_eq_psi_plus` in `sequence/components/bsm.py` is used by `SingleAtomBSM.get` to check whether the stage-1 state was set to `|Psi+>`. In the **density matrix formalism** it compared the stored state against the wrong operator:

```python
elif formalism == DENSITY_MATRIX_FORMALISM:
    d_state = outer(BSM._phi_plus, BSM._psi_plus)   # |Phi+><Psi+|
    return array_equal(state.state, d_state)
```

`outer(phi_plus, psi_plus)` is `|Phi+><Psi+|`, which is **not even a valid density matrix** (not Hermitian, zero trace). Meanwhile `_set_pure_state` stores a pure state `|Psi+>` as `outer(psi_plus, psi_plus) = |Psi+><Psi+|`. The two never match, so the comparison is **always `False`** in the density matrix formalism.

### Impact

In `SingleAtomBSM.get`, stage 2 of Barrett-Kok entanglement generation uses this check to decide the final Bell state:

```python
if _eq_psi_plus(state0, qm.get_active_formalism()) ^ detector_num:
    _set_state_with_fidelity(keys, BSM._psi_minus, ...)
else:
    _set_state_with_fidelity(keys, BSM._psi_plus, ...)
```

Because the check is stuck at `False`, the protocol ignores which detector fired in stage 1 and assigns the **wrong Bell state** to the entangled memories whenever the simulation runs under the density matrix formalism. The ket formalism path (`array_equal(state.state, BSM._psi_plus)`) is correct, which is why the existing tests — all run with the default ket formalism — never exercised the bug.

## Fix

Compare against the correct pure-state density matrix:

```python
d_state = outer(BSM._psi_plus, BSM._psi_plus)
```

## Test plan

- [x] Added `test_eq_psi_plus_density_matrix`, which builds the density matrices of `|Psi+>` and `|Psi->` and asserts `_eq_psi_plus` returns `True` and `False` respectively.
- [x] Existing BSM tests still pass.


tests/components/test_bsm.py
```diff
@@ -23,6 +23,21 @@ def bsm_update(self, src: BSM, msg: dict[str, Any]):
                 raise Exception("invalid update")
 
 
+def test_eq_psi_plus_density_matrix():
+    """_eq_psi_plus must recognize the density matrix of |Psi+> (and reject |Psi->)
+    under the density matrix formalism."""
+    from numpy import outer
```
caitaozhan: Those imports need to be at the top of the file.

caitaozhan (approved):
Looks good!

hayekr (approved):

PR #355, opened 2026-05-21
Fix dropped state collapse in FreeQuantumState.measure_multiple

HasanTasdiq:
## Summary

`FreeQuantumState.measure_multiple` (`sequence/kernel/quantum_state.py`) computes the post-measurement state correctly but then assigns it to attributes that **do not exist** on `FreeQuantumState`:

```python
new_state = new_states[res]
for state in entangled_list:
    state.quantum_state = new_state          # FreeQuantumState has no `quantum_state`
    state.entangled_photons = entangled_list # FreeQuantumState has no `entangled_photons`
return res
```

`FreeQuantumState` stores its vector in `self.state` and its entangled partners in `self.entangled_states` (see the single-qubit `measure` method, which does `s.state = new_state`). Writing to `quantum_state` / `entangled_photons` silently creates **unused** attributes and leaves the real state **uncollapsed**: after `measure_multiple`, the entangled `FreeQuantumState` objects keep their pre-measurement superposition.

### Demonstration

Two qubits each in `|+>`, combined, then measured in the computational basis:

```python
res = FreeQuantumState.measure_multiple(basis, [a, b], rng)   # e.g. res == 2  (=> |10>)
a.state   # buggy: still (0.5, 0.5, 0.5, 0.5); expected (0, 0, 1, 0)
```

This is reached in practice by `Photon.measure_multiple`, used by `PolarizationBSM` and `TimeBinBSM`. The returned index `res` (which drives detector firing) is correct, so detector outcomes are unaffected; but the measured photons' tracked state is left wrong, which is incorrect for any consumer that inspects photon state after a multi-qubit measurement.

## Fix

Assign the collapsed vector to the real fields, mirroring the single-qubit `measure` method:

```python
new_state = tuple(new_states[res])
for qs in entangled_list:
    if qs is not None:
        qs.state = new_state
        qs.entangled_states = entangled_list
return res
```

## Test plan

- [x] Added `test_measure_multiple_collapses_state`, which combines two `|+>` photons, measures in the computational basis, and asserts the shared state collapses to the measured basis vector (and stays shared).
- [x] Existing `test_measure_multiple` and other photon/kernel tests still pass.


sequence/kernel/quantum_state.py
```diff
@@ -361,11 +361,12 @@ def measure_multiple(basis, states, rng: Generator):
...
         # result gives index of the basis vector that will be projected to
         res = rng.choice(possible_results, p=probabilities)
-        # project to new state, then reassign quantum state and entangled photons
-        new_state = new_states[res]
-        for state in entangled_list:
-            state.quantum_state = new_state
-            state.entangled_photons = entangled_list
+        # project to new state, then reassign quantum state and entangled list
+        new_state = tuple(new_states[res])
+        for qs in entangled_list:
```
hayekr, line 366: please call this quantum_state. Avoid abbreviations when able.

tests/components/test_photon.py
```diff
@@ -80,6 +80,27 @@ def test_measure_multiple():
...
+             (complex(0), complex(0), complex(1), complex(0)),
+             (complex(0), complex(0), complex(0), complex(1)))
+
+    res = Photon.measure_multiple(basis, [photon1, photon2], np.random.default_rng(0))
+
+    expected = tuple(complex(1) if i == res else complex(0) for i in range(4))
+    assert tuple(photon1.quantum_state.state) == expected, "state did not collapse to measured basis vector"
+    # both photons must keep sharing the same collapsed state object
+    assert photon1.quantum_state.state is photon2.quantum_state.state
+
```
hayekr, line 102: It would be good to add assertions preventing these invalid class attributes from being used.
```
assert not hasattr(photon1.quantum_state, "quantum_state")
assert not hasattr(photon1.quantum_state, "entangled_photons")
assert not hasattr(photon2.quantum_state, "quantum_state")
assert not hasattr(photon2.quantum_state, "entangled_photons")
```

tests/components/test_photon.py
```diff
@@ -80,6 +80,27 @@ def test_measure_multiple():
...
+
```
hayekr, line 103: You check that state is shared but we should also check that entanglement_states is also shared
`assert photon1.quantum_state.entangled_states is photon2.quantum_state.entangled_states`

hayekr (commented):
Thank you for your contribution. Please see my suggestions before we merge this to RnD.

hayekr (changes requested):

caitaozhan:
Fixed in a new PR #369

PR #357, opened 2026-05-23
feat: Add registry for reservation rule classes

tvk134:
Summary:
- Refactor reservation-generated resource-management rules into individual Rule subclasses.
- Add a ReservationRuleRegistry that maps reservation rule names to Rule constructors.
- Add a DefaultReservationRuleGenerator that uses the registry to construct rules for a reservation.
- Update ResourceManager.generate_load_rules to delegate rule construction while preserving rule scheduling, expiration, and memory-reset behavior.
- Expose ResourceManager.get_reservation_rule_registry() so users can disable or replace individual reservation rules without subclassing the generator.

Testing:
- uv run ruff check sequence/resource_management/resource_manager.py sequence/resource_management/reservation_rule_registry.py tests/resource_management/test_reservation_rule_registry.py
- uv run pytest tests/resource_management -vv
- uv run pytest tests/network_management -vv
- uv run pytest tests/topology -vv
- uv run pytest tests/app -vv
- uv run pytest tests/components -vv
- uv run pytest tests/entanglement_management -vv
- uv run pytest tests/kernel -vv
- uv run pytest tests/qkd -vv
- uv run pytest tests/qlan -vv
- uv run pytest tests/utils -vv

Note:
- Full uv run pytest -vv hangs locally during collection on Windows at 228 collected items, including after a restart and several minutes of waiting.
- Running each top-level test directory individually completes successfully.


sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,286 @@
...
+            condition_args = {"memory_indices": memory_indices[reservation.memory_size:]}
+
+        action_args = {
+            "mid": owner.map_to_middle_node[path[index + 1]],
+            "path": path,
+            "index": index,
+            "name": owner.name,
+            "reservation": reservation,
+        }
+        super().__init__(10, eg_rule_action_request, eg_rule_condition, action_args, condition_args)
```
caitaozhan: 10 is the priority. Needs to pass in as an argument in the __init__()

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,286 @@
...
+    """Entanglement-generation await rule for non-left-end nodes."""
+
+    def __init__(self, owner, path, reservation, memory_indices, index) -> None:
+        condition_args = {"memory_indices": memory_indices[:reservation.memory_size]}
+        action_args = {
+            "mid": owner.map_to_middle_node[path[index - 1]],
+            "path": path,
+            "index": index,
+        }
+        super().__init__(10, eg_rule_action_await, eg_rule_condition, action_args, condition_args)
```
hayekr: These should be parameters in the class

tvk134:
Pushed the discussed updates.

Summary:
- Renamed DefaultReservationRuleGenerator to ReservationRuleGenerator.
- Removed the custom registry injection path; the generator now owns its registry initialized with the default reservation rule constructors.
- Kept ResourceManager.get_reservation_rule_registry() as the configuration hook for replacing/disabling individual reservation-generated rules.
- Made reservation rule priority an __init__ parameter with the existing default value of 10.
- Added a generator-level test verifying that a registered replacement rule is actually produced by create_rules(...), not just stored in the registry.
- Cleaned up the reservation rule registry tests.

Validation:
- uv run pytest tests/resource_management/test_reservation_rule_registry.py
  - All 4 passed
- uv run pytest tests/resource_management
  - All 16 passed

hayekr:
Please resolve your conflict.

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,340 @@
...
+        priority: int = 10,
+    ) -> None:
+        left, right = _get_swapping_neighbors(owner.name, path)
+        condition_args = {
+            "memory_indices": memory_indices,
+            "left": left,
+            "right": right,
+            "fidelity": reservation.fidelity,
+        }
+        super().__init__(priority, es_rule_action_A, es_rule_condition_A, {}, condition_args)
```
Tw1ZZLER: Why is `action_args` empty here when the original rule creation passes `swapping_success_prob` and `swapping_degradation`? this is a major regression in behavior, please fix.

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,340 @@
...
+    EG_AWAIT: EGAwaitRule,
+    EG_REQUEST: EGRequestRule,
+    EP_REQUEST: EPRequestRule,
+    EP_AWAIT: EPAwaitRule,
+    ES_B_END: ESBEndRule,
+    ES_A: ESARule,
+    ES_B: ESBRule,
+}
+
+DEFAULT_RESERVATION_RULE_ORDER: Final[tuple[str, ...]] = (
```
Tw1ZZLER: This is never referenced

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,340 @@
...
+            raise ValueError(msg)
+
+
+class ReservationRuleGenerator:
+    """Rule generator for reservation-based resource-management rules."""
+
+    def __init__(self) -> None:
+        self.registry = ReservationRuleRegistry(DEFAULT_RESERVATION_RULE_CONSTRUCTORS)
+
+    def create_rules(self, owner, path, reservation, memory_indices, index) -> list[Rule]:
```
Tw1ZZLER: What was the purpose of defining `DEFAULT_RESERVATION_RULE_ORDER` if you're just going to hardcode the order again in this function?

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,340 @@
...
+        """Build a rule by name.
+
+        Returns None when the rule is disabled.
+        """
+        constructor = self.get(name)
+        if constructor is None:
+            return None
+        return constructor(*args, **kwargs)
+
+    def copy(self) -> "ReservationRuleRegistry":
```
Tw1ZZLER: This method has no caller

sequence/resource_management/resource_manager.py
Tw1ZZLER: This is initalized and then immediately overwritten by line 167-173

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,340 @@
...
+    EP_AWAIT,
+    ES_B_END,
+    ES_A,
+    ES_B,
+)
+
+ReservationRuleConstructor = type[Rule]
+
+
+class EGAwaitRule(Rule):
```
Tw1ZZLER: All of these 7 new Rule classes are almost identical and are copy-pasted, with the only difference being the condition and action args passed to the constructors. Why not use a single table of (name, predicate, builder) instead of defining a bunch of classes that try to be uniform in their constructors?

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,340 @@
...
+class EPRequestRule(Rule):
+    """Entanglement-purification request rule."""
+
+    def __init__(
+        self,
+        _owner,
+        _path,
+        reservation,
+        memory_indices,
+        _index,
```
Tw1ZZLER: The underscore parameters tell me that the "uniform" API you are trying to create is just wrong for half of the rules...

sequence/resource_management/reservation_rule_registry.py
Tw1ZZLER: This control flow feels identical to the control flow from the original method that you deleted. What was the point of creating the separate classes if the control flow is the exact same and not iterated or improved.

This method also contributes to a lower level of extensibility. The user has no control over when rules are created, there is just these hardcoded if statements. You can't change when the rules are created without subclassing ReservationRuleGenerator and rewriting the create_rules method.

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,340 @@
...
+        return ReservationRuleRegistry(self._constructors.copy())
+
+    @staticmethod
+    def _validate_name(name: str) -> None:
+        if name not in RESERVATION_RULE_NAMES:
+            msg = f"Unknown reservation rule name: {name}"
+            raise ValueError(msg)
+
+
+class ReservationRuleGenerator:
```
Tw1ZZLER, line 299: Your PR description reads "Added DefaultReservationRuleGenerator" but this doesn't match that

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,340 @@
...
+        if constructor is None:
+            return None
+        return constructor(*args, **kwargs)
+
+    def copy(self) -> "ReservationRuleRegistry":
+        """Return a shallow copy of this registry."""
+        return ReservationRuleRegistry(self._constructors.copy())
+
+    @staticmethod
+    def _validate_name(name: str) -> None:
```
Tw1ZZLER, line 281: So you validate a name to be registered, by checking the name against a final value that cannot be changed? How are any new rules registered that aren't hard-coded? This seems like it intends on the user editing the source to add any new reservation rules.

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,340 @@
...
+
+EG_AWAIT: Final[str] = "eg_await"
+EG_REQUEST: Final[str] = "eg_request"
+EP_REQUEST: Final[str] = "ep_request"
+EP_AWAIT: Final[str] = "ep_await"
+ES_B_END: Final[str] = "es_b_end"
+ES_A: Final[str] = "es_a"
+ES_B: Final[str] = "es_b"
+
+RESERVATION_RULE_NAMES: Final[tuple[str, ...]] = (
```
Tw1ZZLER, line 49: This is constant and cannot be changed, meaning no new reservation rules can be added without editing the source.

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,340 @@
+"""Registry utilities for reservation-generated resource-management rules.
+
+This module defines reservation-specific Rule subclasses and a registry that
+maps reservation rule names to Rule constructors. The default resource-management
+rule generator uses this registry to construct the built-in rules, while users
+may replace or disable individual rules without subclassing the generator.
```
Tw1ZZLER: Okay, technically true, but in practice would rarely ever be the case. "Replace" would only mean swapping one of the existing seven constructors, which is not very extensible. If you want to change scheduling at all you still need to subclass the generator, since the order in `create_rules` is hardcoded. This really overstates what the registry can do

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,340 @@
...
+    def __init__(self, constructors: dict[str, ReservationRuleConstructor] | None = None) -> None:
+        self._constructors: dict[str, ReservationRuleConstructor] = {}
+        if constructors is not None:
+            for name, constructor in constructors.items():
+                self.register(name, constructor)
+
+    def register(self, name: str, constructor: ReservationRuleConstructor) -> None:
+        """Register or replace a reservation Rule constructor."""
+        self._validate_name(name)
+        if not issubclass(constructor, Rule):
```
Tw1ZZLER: Only checks that the passed constructor is a subclass of Rule, not the constructor accepts the proper arguments. Any typo in a custom constructor would result in a runtime crash on the first reservation, not on registration. This suggests more problems with the "uniform" API.

Tw1ZZLER (changes requested):
This PR mostly reads like a prompt-driven refactor, and not an intentional redesign of how a good reservation rule registry should work. I would seriously spend more time thinking about how the API should be re-engineered for this.

I do think the addition of `_get_swapping_neighbors` was a great change! The introduction of this method makes the readability much much better and gives the loop a good name.

However, extensibility is **pretty overstated**, and in practice most users would still have to override a lot of the methods and existing framework. Please see comments for other thoughts.

I really think the motivation behind this change is a good one, but the execution does not feel very thought out in terms of engineering and design.

Tw1ZZLER:
The PR's scope is not clear given the description of the PR and the code itself. It appears that you are trying to partially implement a completely generalized solution for a rule registry, that allows users to add rules and change the order of rules however they see fit. 

If you are going for a simpler system that only replaces existing rules with custom args, then I would suggest using a simple table of **(name, predicate, builder)** instead of classes for each rules, and make it explicit that the registry only allows the user to replace the reservation rules with custom ones. 

I think there was a discrepancy in the description and the code itself versus the purpose and scope. Overall I think the implementation would make a good contribution towards SeQUeNCe, but it is a little over-engineered for the simple "replace rules easily" scope. Try and make it simpler and narrow the scope and it will be a very good addition I think :slightly_smiling_face: 

Ignore the comments from my review until you narrow the scope. I will resolve comments that fall outside of the new scope or are outdated :100:

tvk134:
Thanks @Tw1ZZLER , this clarification helped. I revised the PR around the narrower intended scope: this is now explicitly a fixed-scope replacement mechanism for SeQUeNCe’s existing reservation-generated rule slots, not a generalized registry for adding arbitrary reservation phases or changing scheduling order.

The implementation now uses a static table of (name, predicate, builder) specs instead of separate Rule subclasses. The registry allows replacing or disabling builders for those existing rule slots, while the rule creation order and predicates remain fixed. I also preserved the ES swapping success/degradation action arguments, removed the dead/misleading API pieces from the earlier version, and updated the tests for the narrowed API.

I ran the pytest suite locally and all tests pass. Please let me know if I missed anything from the intended scope or if there are any additional changes you’d like me to make.

sequence/resource_management/resource_manager.py
```diff
@@ -148,10 +138,15 @@ def __init__(self, owner: QuantumRouter, memory_array_name: str):
         self.pending_protocols = [] # Protocols that are requesting remote resource
         self.waiting_protocols = [] # Protocols that are waiting request from remote resource
         self.memory_to_protocol_map = {}
+        self.rule_generator = ReservationRuleGenerator()
 
     def init(self):
         pass
 
+    def get_reservation_rule_registry(self) -> ReservationRuleRegistry:
+        """Return the registry used to replace or disable reservation rule builders."""
```
Tw1ZZLER: Can you add a usage example here showing how to customize a rule? It will help for documentation

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,303 @@
...
+    condition_args = {
+        "memory_indices": context.memory_indices,
+        "target_remote": target_remote,
+        "fidelity": context.reservation.fidelity,
+    }
+    return Rule(10, es_rule_action_B, es_rule_condition_B_end, {}, condition_args)
+
+
+def _get_swapping_neighbors(owner_name: str, path: list[str]) -> tuple[str, str]:
+    """Return the left and right neighbors used for middle-node swapping rules."""
```
Tw1ZZLER: Can you explain the algorithm used here just a bit more?

Tw1ZZLER (dismissed):
Looks great!! You did an amazing job implementing my idea, probably much better than I would have done so. `create_rules()` is now such a clean method it makes me happy just looking at it! Definitely a massive improvement over what it used to be before this PR, and gives the user more customizability for sure. **APPROVE!**

I added 2 comments for adding additional documentation in some spots that would be nice, but not blocking the merge.

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+def _build_eg_await_rule(context: ReservationRuleContext) -> Rule:
+    condition_args = {"memory_indices": context.memory_indices[:context.reservation.memory_size]}
+    action_args = {
+        "mid": context.owner.map_to_middle_node[context.path[context.index - 1]],
+        "path": context.path,
+        "index": context.index,
+    }
+    return Rule(10, eg_rule_action_await, eg_rule_condition, action_args, condition_args)
```
caitaozhan: Priority cannot be fixed at `10`. Can it be passed in as an argument?

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+def _build_eg_request_rule(context: ReservationRuleContext) -> Rule:
+    if context.index == 0:
+        condition_args = {"memory_indices": context.memory_indices[:context.reservation.memory_size]}
+    else:
+        condition_args = {"memory_indices": context.memory_indices[context.reservation.memory_size:]}
+
+    action_args = {
```
caitaozhan, line 125: What if I want to change the `action_args`? 

See an example of customizing the `action_args` by adding `"memo_type"` https://github.com/haydenmllr1317/heterogenous/blob/be30fa92a4e101a6ac64609d4232889d0fc430ca/resource_manager.py#L164

caitaozhan (changes requested):

sequence/resource_management/reservation_rule_registry.py
```diff
@@ -0,0 +1,303 @@
...
+            raise TypeError(msg) from exc
+
+
+class ReservationRuleGenerator:
+    """Rule generator for reservation-based resource-management rules."""
+
+    def __init__(self) -> None:
+        self.registry = ReservationRuleRegistry()
+
+    def create_rules(self, owner, path, reservation, memory_indices, index) -> list[Rule]:
```
caitaozhan: Typing is missing in `create_rules`'s arguments. Docstring is incomplete. Need an `Args` section and a `Returns` section.

This comment applies to all other places.

caitaozhan (changes requested):

sequence/resource_management/reservation_rule_registry.py
tvk134: Addressed. I added explicit typing to the create_rules() arguments and expanded its docstring with Args: and Returns: sections.

sequence/resource_management/reservation_rule_registry.py
tvk134: Addressed. The linked example customizes the generated EG rule by adding extra action_args such as memo_type. I documented that this can be done by replacing the relevant built-in rule builder and returning a full Rule with customized action_args. I kept this at the builder level rather than adding a separate action_args mutation hook, since the registry abstraction replaces/disables reservation-generated rule builders.

sequence/resource_management/reservation_rule_registry.py
tvk134: Addressed. Rule priority is now carried in ReservationRuleContext, ReservationRuleGenerator.create_rules(..., priority=10) accepts a priority argument, and ResourceManager.generate_load_rules(..., rule_priority=10) passes it through. The default remains 10 to preserve existing behavior, but generated default rules now use context.priority instead of hardcoding 10.

sequence/resource_management/resource_manager.py
tvk134: Addressed in the latest commit. I added documentation to ResourceManager.get_reservation_rule_registry() showing that replacement builders receive a ReservationRuleContext and return the full Rule, so callers can customize priority, condition args, and action args for an existing reservation rule slot.

sequence/resource_management/reservation_rule_registry.py
tvk134: Addressed. I expanded the _get_swapping_neighbors() docstring to explain the path-reduction logic used for middle-node swapping rules: the path is reduced layer by layer until the current node is at an odd index, then its immediate reduced-path neighbors become the left/right swapping partners.

tvk134:
Thanks for the comments and feedback. I pushed a follow-up commit addressing the review points: added a usage example for replacing a built-in rule builder, documented how replacement builders can customize action_args, expanded the swapping-neighbor docstring, threaded configurable rule priority through the generator/resource-manager path while preserving the default priority of 10, and added regression coverage for the priority path. Tests pass locally.

tvk134:
Updated the default reservation-rule priorities to preserve SeQUeNCe’s existing ordering: EG uses the base priority, EP uses base + 10, and ES uses base + 20. Also added regression coverage for both default and custom base priorities; the full resource-management test suite passes.

hayekr:
@caitaozhan Can you take a pass on this. It should be ready to go.

PR #364, opened 2026-06-05
Error handling for missing quantum channels

Tw1ZZLER:
Error handling for the `send_qubit` function so that when a node attempts to send a qubit to another node, if there is no quantum channel between the two, a `ValueError` gets raised, as opposed to Python crashing.

> First PR! :grin: :grin: :grin:


sequence/topology/node.py
```diff
@@ -172,9 +172,15 @@ def schedule_qubit(self, dst: str, min_time: int) -> int:
...
-        """Interface for quantum channel `transmit` method."""
+        """Interface for quantum channel `transmit` method.
 
-        self.qchannels[dst].transmit(qubit, self)
+        If no quantum channel is available to transmit qubit, a ValueError gets raised.
+        """
+
+        if dst not in self.qchannels:
+            raise ValueError(f"No available quantum channel to send qubit from sending node \"{self.name}\" to receiving node \"{dst}\"")
+        else:
```
hayekr: The else line is not needed. 
*.transmit() will not be called if the ValueError is raised, thus `else:` is redundant.

sequence/topology/node.py
```diff
@@ -172,9 +172,15 @@ def schedule_qubit(self, dst: str, min_time: int) -> int:
...
     def send_qubit(self, dst: str, qubit) -> None:
-        """Interface for quantum channel `transmit` method."""
+        """Interface for quantum channel `transmit` method.
 
-        self.qchannels[dst].transmit(qubit, self)
+        If no quantum channel is available to transmit qubit, a ValueError gets raised.
+        """
+
+        if dst not in self.qchannels:
+            raise ValueError(f"No available quantum channel to send qubit from sending node \"{self.name}\" to receiving node \"{dst}\"")
```
hayekr: Call the __repr__ on self.name instead of adding the quotes with escape yourself. Example {self.name!r}

hayekr (changes requested):

hayekr (approved):
lgtm

Issue #289, opened 2026-02-04
Update documentation


hayekr:
Documentation updates, i.e., tutorial website, usage, installation.

PR #366, opened 2026-06-08
Update tutorial and README

caitaozhan:
* The three-minute tutorial was extensively rewritten: the Python example was simplified and clarified,
* The installation instructions in `README.md` were updated to provide clear, step-by-step guidance for both `pip` and `uv` workflows, with explanations of editable installs and environment management. 
* The index page was fixed to reference the correct tutorial file name.
* The table of papers using SeQUeNCe was updated with new entries and code links


caitaozhan:
Link this PR to the issue https://github.com/sequence-toolbox/SeQUeNCe/issues/289

docs/source/about/requirements/prerequisite.md
hayekr, line 58: Lets make sure we cut this. uv auto does editable mode and we no longer use setuptools

README.md
```diff
@@ -44,9 +44,27 @@ pip install sequence
 ```
 
 ### Development Environment Setup     
-If you wish to modify the source code, use an editable installation with [uv](https://docs.astral.sh/uv/):
+If you wish to modify the source code, use an editable installation with either pip or [uv](https://docs.astral.sh/uv/):
```
hayekr, line 47: We should not support editable pip installers

hayekr (changes requested):
Main comment is we need to remove pip editable install now that we have uv managing this.

README.md
caitaozhan: I still use conda and pip (these are 10+ year habits that won't change easily). The majority of the people out there won't be using uv (either don't know it or just don't want to use it after knowing it). This package is managed by uv which is fine. But a random developer out there might not want to use it.

Therefore, we should keep both ways.

docs/source/about/requirements/prerequisite.md
caitaozhan: NO. They are not mutually exclusive. pip and uv can exist at the same time, depending on what you like to use.

hayekr (approved):

PR #372, opened 2026-06-17
Add link length and attenuation to graph building CLI

hayekr:
Exposes the QC link length and attenuation parameters to the CLI. They existed previously in the graph creation but as kwargs with default values. They are now exposed to the user.


caitaozhan (approved):
Looks good!

PR #376, opened 2026-06-22
feat: Metrics tracking system for entanglement management 

### LINKED ISSUE #368: Centralized metrics tracking system
This is a living document that will be iterated on. See previous edits for previous iterations.
---

In order to record metrics into SeQUeNCe, we currently need to monkey-patch the EG, EP, and ES protocols to record the metrics that we need. [An example of this monkey-patching is shown here](https://github.com/sequence-toolbox/entanglement_purification_experiment/blob/6c49f649bdacfeebefcd85c91e9902151d775368/ep_experiment.py#L20-L133). This is obviously not ideal for quickly designing experiments, as the designer must worry about which protocols within sequence they need to patch and override to track the metrics. 

The solution to this problem would be a **centralized metrics tracking system built into SeQUeNCe** that tracks entanglement events such as:

- `eg_attempt`: Entanglement generation attempt
- `eg_success`: Successful entanglement generation attempt
- `ep_attempt`: Purification attempt
- `ep_success`: Successful purification attempt

This list is by no means final, and I will definitely add to it as I come up with more metrics to track. This was mostly based on the example given above. If there are any metrics you think are missing, please feel free to comment them! (or edit the issue and add them if you would prefer)

## Proposed Solution

The proposed way to implement this would be a global `metrics` module within SeQUeNCe that has attributes `time_provider` and `storage`. The `time_provider` attribute would typically be registered as the `Timeline` of the application, while `storage` would likely be a new, custom storage class that I will discuss a little bit later.

`metrics` would be created and managed automatically by SeQUeNCe, and would have many default metrics that it tracks automatically. `metrics` is a **global registry module** imported simply as `from sequence import metrics`. The user then calls the front-facing methods such as `metrics.configure()` and `metrics.enable()`

Entanglement generation, purification, and swapping protocols would have to have their methods modified to include calls to `metrics.record()` to record their metrics. It would be similar to how they are currently overridden, they would just live within those methods themselves.

### Record Interface

The actual recording of the metrics would be done via a `metrics.record()` method. Here is an example of how it could be implemented in the `_entanglement_succeed()` method of `SingleHeraldedA` protocol.

```python
class SingleHeraldedA(Protocol):
    def _entanglement_succeed(self):
        log.logger.info(f'{self.owner.name} successful entanglement of memory {self.memory}')

        # Added call to metrics.record here
        metrics.record(
            # Pass Event Type Enum
            event_type=metrics.EG_SUCCESS,

            # Store name of node in which event occurred
            node=self.owner.name,

            # Metrics that we want to record
            metrics={'mem_fidelity'}
        )

        self.memory.entangled_memory['node_id'] = self.remote_node_name
        self.memory.entangled_memory['memo_id'] = self.remote_memo_id
        self.memory.fidelity = self.raw_fidelity

        self.update_resource_manager(self.memory, MemoryInfo.ENTANGLED)
```

There are three arguments:
- `event_type`: An enum that describes on what event this metric was recorded, can also be used to increment counts of events (ex. this is the 5432th entanglement success)
- `node`: Name of node in which event occurred
- `**kwargs`: Flexible dict of extra metrics that can be passed on a single event

### Multithreading

`metrics` and all of it's functions wouid run on a separate thread from the main SeQUeNCe simulation in order to minimize overhead. Once the protocols call `metrics.record()`, all the work is done on the `metrics` thread. 

### Storage Class

An abstract `BaseMetricsStorage` class will be created that lists common methods for metric storage methods to have. Two proposed storage methods are `SQLiteStorage` and `InMemoryStorage`. `InMemoryStorage` would likely be the default, and would store the metrics within the application's memory at runtime. `SQLiteStorage` would be a more robust solution that would store metrics within a SQLite database. This would take pressure off the memory and keep the data stored on-disk in case of application crashes or any other scenario, and would likely be the only option for massive experiments.

### Enable

It is likely that `metrics` could cause some kind of performance overhead, especially with large experiments. `metrics` will not run or do anything until the user has ran `metrics.enable()`, to minimize any performance overhead. The user must enable specific recording instances such as `eg_success`, `eg_attempt`, etc. to be enabled before any metrics are actually recorded. A future `metrics.enable_all()` function could be implemented that simply enables `metrics` globally and instantly enables all instances of `metrics.record()`. For now, the individual `enable()` implementation will be used.

Example usage:

```python
metrics.enable([metrics.EG_SUCCESS, metrics.EG_ATTEMPT])
```

### User Configuration

All the user would have to do is optionally call a `configure()` function to change the storage type (sqlite or in-memory), and if using sqlite then pass a `file_path` argument for the `SQLiteStorage` to use. For example, if we go with the **global registry module** method of implementing metrics:

```python
from sequence import metrics
metrics.configure(storage_type="sqlite", file_path="cool_simulation_metrics.db")
```

If a user wants to implement their own protocol and record it's metrics, they would use the `record()` method in their own protocol as shown above in Record Interface.

### Diagram

Image

---

Feel free to let me know your thoughts 🙂

Tw1ZZLER:
Continuation of implementation of #368 and direct continuation of development after #374.
---

This PR implements a more generalized central metrics tracking system for SeQUeNCe. The module has been split up until multiple modules:
- `metrics/__init__.py`: The front-facing API methods that are used to interact with the metrics module. Includes `enable`, `record`, `configure`, `collect_trial_metrics`, and `aggregate_trial_metrics`. 
- `event_types.py`: Defines methods for creating and handling event types. Event types are what is used by the `record` method to record a metric.
- `builtins.py`: Built-in event types and metrics. Used for abstraction and categorizing the different types of metrics used throughout SeQUeNCe. Also gives an example of how a user could create their own metrics for their own experiments!
- `metrics.py`: Mostly defines a Metric abstract class and the types of Metrics that are available. Each metric includes `collect` method that collects the information from those metrics, and this is typically used with the `collect_trial_metrics` method.
- `registry.py`: Metric registry that provides convenient methods for registering and creating new metrics. All Metric types are stored as a list `_metrics` in this module.
- `storage.py`: Mostly unchanged from #374, just all storage methods have been moved to their own module, along with the the time provider classes.

## Metric Superclass

Metrics have been generalized into certain metric types, that change the behavior of certain methods. `Metric` is now an abstract superclass of all the other metric types, which are implemented as data classes.

### Abstract Properties

Metrics requires the abstract properties `event_types` and `output_keys` to be set: 
- `event_types` is an immutable set of the EventTypes that the metric reacts to during `record`. 
- `output_keys` is an immutable set of the keys that the `collect()` method produces in it's output dictionary. Useful for aggregation of data. `collect()` is discussed in the next section.

### Collect Method

Metric also requires an abstract method in the form of `collect()`. The `collect()` method is to generalize the global `collect_trial_metrics()` method, which now just calls the `collect()` method of each metric and puts them all into one list. `collect()` is implemented differently and outputs different sets of information based on the metric. 

- `CounterPairMetric` collects the failures, successes, and success rate of the passed owner. This helps generalize the successes and failures that all the entanglement protocols produce.
- `LastValueMetric` collects the last scalar field value from matching events. This helps generalize values like the throughput of an app.
- `EventFieldListMetric` collects a list of field values matching a certain event type. This helps generalize giant lists of fidelities, like purified fidelities over time.
- `DeliveryTimeMetric` collects the time to deliver N purified pairs relative to reservation start. This helps generalized values like purified delivery of the entanglement purification protocol (RequestApp, etc.).

### On Record Method

`on_record()` is an optional method that is called when the metric is recorded. It is only used by `CounterPairMetric`, where it updates the internal success and failure counters and updates the success rate.

## Collect Context

`CollectContext` is a simple data class that lays out a template context to pass to the `collect()` method of each metric. It also tells the `collect()` method what it can expect when it gets called. It is only used by the `collect_trial_metrics()` method. **This need some more work to be properly generalized or at least easily extended by the user.**

## Collect Trial Metrics Method

The `collect_trial_metrics` method is now more generalized for all types of metrics. All it does now is iterate through the list of metrics and calls the `collect()` method of each metric. It uses the created `CollectContext` object with the passed arguments, to pass those arguments through to the metrics `collect()` methods. **However, these arguments are currently specific to certain protocols and are not currently generalized.**


Tw1ZZLER:
Metrics support for recording failures, successes, and fidelities for all of entanglement management has been completed. Future work on the other layers of SeQUeNCe (like network management) will be in a different PR. Ready for review :)

sequence/entanglement_management/generation/generation_base.py
```diff
@@ -180,6 +180,7 @@ def _entanglement_succeed(self):
     def _entanglement_fail(self):
         for event in self.scheduled_events:
             self.owner.timeline.remove_event(event)
+        metrics.record(metrics.EG_FAILURE, self.owner.name, initial_fidelity=self.fidelity)
```
caitaozhan: why is the argument `initial_fidelity` in recording EG_FAILURE and `fidelity` in recording EG_SUCCESS.

Also, the first argument is `metrics.EG_FAILURE`. It would be more clear if the argument is something like `EventType.EG_FAILURE`

Also, entanglement generation is a protocol involving two nodes. So, I think recording just the `self.owner.name` is not enough.

sequence/utils/metrics/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+    register_metric,
+    reset_metrics,
+    unregister_metric,
+)
+from .storage import InMemoryStorage, SystemTimeProvider, TimeProvider
+
+# Built-in event type aliases
+EG_FAILURE = builtins.EG_FAILURE
+EG_SUCCESS = builtins.EG_SUCCESS
+THROUGHPUT = builtins.THROUGHPUT
```
caitaozhan: `THROUGHPUT` is not an event. The EG, EP, ES success/failures are events.

sequence/utils/metrics/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+EP_FAILURE = builtins.EP_FAILURE
+EP_SUCCESS = builtins.EP_SUCCESS
+PURIFIED_DELIVERY = builtins.PURIFIED_DELIVERY
```
caitaozhan: I frequently turn off purification and do not use it at all. So purification is not a necessity. I do not use purification because it (often) does more harm than good in the simulations.

Changing `PURIFIED_DELIVERY` to just `DELIVERY` may be better.

sequence/utils/metrics/registry.py
```diff
@@ -0,0 +1,55 @@
...
+_counter_pairs: dict[str, CounterPairMetric] = {}
+
+
+def register_metric(metric: Metric) -> None:
+    """Register a metric. Raises if output keys collide with existing metrics."""
+    new_keys = metric.output_keys
+    for existing in _metrics:
+        overlap = existing.output_keys & new_keys
+        if overlap:
+            raise ValueError(
```
caitaozhan:
```
s = '            raise ValueError(f"Metric output keys {sorted(overlap)} already registered.")'
len(s) = 89
```
The length is 89, smaller than 120. So it should be in one line, rather than three. 120 characters is the line length limit Robert and I agreed on. Let's apply the 120-character limit to other places as well.

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,207 @@
...
+    @abstractmethod
+    def event_types(self) -> frozenset[EventType]:
+        """Event types this metric reacts to during recording."""
+
+    @property
+    @abstractmethod
+    def output_keys(self) -> frozenset[str]:
+        """Keys produced by collect()."""
+
+    def on_record(
```
caitaozhan: the `on_record` function fits under the 120-char limit too. It can be written in one line, just like the rest of the repository. Follow the convention.

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,207 @@
...
+            f"{self.prefix}_success_rate": self.success_rate(ctx.owner_name),
+        }
+
+    def reset(self) -> None:
+        self._failures.clear()
+        self._successes.clear()
+
+
+@dataclass
+class LastValueMetric(Metric):
```
caitaozhan: The naming `LastValueMetric` is not intuitive and hard to understand.

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,207 @@
...
+    @abstractmethod
+    def collect(self, ctx: CollectContext) -> dict[str, Any]:
+        """Return trial result keys and values for this metric."""
+
+    def reset(self) -> None:
+        """Clear per-trial metric state."""
+
+
+@dataclass
+class CounterPairMetric(Metric):
```
caitaozhan: What about `CounterMetric` -- no pair?

In the RequestApp, there is an attribute `memory_counter`

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,207 @@
...
+            for record in ctx.storage.get_by_owner(ctx.owner_name)
+            if record["event_type"] is self.event
+        ]
+        if not records:
+            return {self.key: float("nan")}
+        return {self.key: records[-1][self.field]}
+
+
+@dataclass
+class EventFieldListMetric(Metric):
```
caitaozhan: This naming `EventFieldListMetric` is also odd.

In general, the metrics have three types: counter, time, and fidelity. Counter/time is the rate, which could be a fourth metric. 

Whether it is a scalar or a list doesn't matter. A scalar is a list of one element.

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,207 @@
...
+            record[self.field]
+            for record in ctx.storage.get_by_owner(ctx.owner_name)
+            if record["event_type"] is self.event and self.field in record
+        ]
+        return {self.key: values}
+
+
+@dataclass
+class DeliveryTimeMetric(Metric):
+    """Time to deliver N purified pairs relative to reservation start."""
```
caitaozhan: DeliveryTimeMetric should not be restricted to purified pairs.

caitaozhan (commented):
I left some comments. Please take a look.

sequence/utils/metrics/registry.py
Tw1ZZLER: Ruff's default line length when formatting is 88. I will edit the pyproject.toml file to make the global project line length for Ruff to be 120, and that should fix any future formatting issues (assuming everyone is using Ruff), and I will reformat these files

sequence/utils/metrics/metric_types.py
Tw1ZZLER: Fixed by changing Ruff line length to 120 and reformatting, also had to remove trailing commas.

sequence/utils/metrics/metric_types.py
Tw1ZZLER: Good suggestion! I renamed the metric types to be a little more intuitive

sequence/utils/metrics/metric_types.py
Tw1ZZLER: I definitely agree - I renamed accordingly

sequence/utils/metrics/metric_types.py
Tw1ZZLER: Renamed to `RateMetric` following your other suggestion

sequence/entanglement_management/generation/generation_base.py
Tw1ZZLER: Renamed that `initial_fidelity` argument to just `fidelity`. Added new namespace `EventTypes` for both the built-in and registered event types. I also added a new argument for the remote node to each of the metrics that involve two nodes, like generation

sequence/utils/metrics/__init__.py
Tw1ZZLER: THROUGHPUT event removed - I instead added it to the collection context on `collect_trial_metrics`. Will eventually implement some simpler API for the `RateMetric`s to be collected when `collect_trial_metrics` is called, but this works for now

sequence/utils/metrics/__init__.py
Tw1ZZLER: changed to just `DELIVERY`

sequence/utils/metrics/metric_types.py
Tw1ZZLER: yes, that makes more sense. I edited the method comment

sequence/utils/metrics/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+from __future__ import annotations
+
+import math
+from statistics import mean, stdev
+from typing import Any
+
+from . import builtins
+from .event_types import (
+    EventType,
+    EventTypes,
```
caitaozhan, line 19: Are `EventType` and `EventTypes` both needed? Looks a bit odd.

sequence/utils/metrics/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+    get_event_type,
+    list_event_types,
+    register_event_type,
+)
+from .metric_types import (
+    CollectContext,
+    CounterMetric,
```
caitaozhan, line 26: In my VSCode, lots of the imports here are dimmed -- they are not used in this `__init__`. Not a serious problem, but I always avoid dimmed stuff. Is importing everything in the `__init__` needed here?

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def success_rate(self, owner_name: str) -> float:
+        failures = self.failures(owner_name)
+        successes = self.successes(owner_name)
+        attempts = failures + successes
+        if attempts == 0:
+            return 0.0
+        return successes / attempts
+
+    def on_record(self, event_type: EventType, owner_name: str, kwargs: dict[str, Any]) -> None:
```
caitaozhan, line 155: Lacks a docstring. Other places where there is a docstring lack `Args` or `Returns`. 

A complete docstring goes like:
```
def function():
   """Description

    Args:

    Returns: 
        (Note: if nothing to return, then this could skip)
   """
```

caitaozhan (commented):
Only some minor comments left!

sequence/utils/metrics/__init__.py
Tw1ZZLER: Me and Robert discussed this quite a bit, `EventTypes` is the "public-facing" API that will be used a majority of the time, while `EventType` is only there for typing within the metrics module. I think for now it should be fine and there should not be too much confusion, but a better implementation of the `EventTypes` namespace is planned

sequence/utils/metrics/__init__.py
Tw1ZZLER: These imports are what is used when other modules of SeQUeNCe, or the user, imports anything from metrics such as `EventTypes`. My type checker does the same thing, so it does bug me a little, but when I remove them the tests all fail because they can't import anything.

sequence/utils/metrics/metric_types.py
Tw1ZZLER: Added all the docstrings in the requested format

sequence/utils/metrics/storage.py
```diff
@@ -0,0 +1,78 @@
...
+
+from .event_types import EventType
+
+
+class TimeProvider(Protocol):
+    """Protocol for objects that supply a timestamp for recorded events."""
+
+    def now(self) -> int:
+        """Return the current timestamp in picoseconds."""
+        ...
```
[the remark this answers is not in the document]
hayekr: @Tw1ZZLER Please resolve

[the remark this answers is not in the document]
Tw1ZZLER: Fixed, replaced with a `NotImplementedError` raise

sequence/utils/metrics/storage.py
```diff
@@ -0,0 +1,78 @@
...
+
+class TimeProvider(Protocol):
+    """Protocol for objects that supply a timestamp for recorded events."""
+
+    def now(self) -> int:
+        """Return the current timestamp in picoseconds."""
+        raise NotImplementedError
+
+
+class SystemTimeProvider:
```
hayekr: We should never use system time as the time provider.

sequence/utils/metrics/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+_enabled = False
+_enabled_events: set[EventType] = set()
+storage = InMemoryStorage()
+_system_time_provider = SystemTimeProvider()
+time_provider: TimeProvider = _system_time_provider
+
+
+def register_time_provider(provider: TimeProvider) -> None:
```
hayekr: Just use timeline as the type. No need for the abstraction.

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+@dataclass
+class CollectContext:
+    """Context passed to metrics when collecting trial results.
+
+    Attributes:
+        owner_name: Node name for counter and fidelity metrics.
+        storage: In-memory store of recorded events for the trial.
+        delivery_owner: Node name for delivery-time metrics; defaults to ``owner_name``.
```
hayekr: It currently defaults to `None` not `owner_name`

hayekr: also inline code entries are `` not ````

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,281 @@
...
+        reservation_start_time: Simulation time when the reservation started (ps).
+        throughput: Application throughput supplied at collection time.
+    """
+
+    owner_name: str
+    storage: InMemoryStorage
+    delivery_owner: str | None = None
+    target_pairs: int = 500
+    reservation_start_time: int | None = None
+    throughput: float | None = None
```
hayekr, line 31: Remove throughput

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+class Metric(ABC):
```
hayekr, line 34: Unimplemented methods should as least have the `pass` instruction

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,281 @@
...
+    @property
+    @abstractmethod
+    def output_keys(self) -> frozenset[str]:
+        """Keys produced by ``collect()``.
+
+        Returns:
+            Frozen set of keys written into per-trial result dictionaries.
+        """
+
+    def on_record(self, event_type: EventType, owner_name: str, kwargs: dict[str, Any]) -> None:
```
hayekr, line 55: Specify that these are optional in docstring

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,281 @@
...
+        delivery_owner: Node name for delivery-time metrics; defaults to ``owner_name``.
+        target_pairs: Number of delivered pairs required to compute delivery time.
+        reservation_start_time: Simulation time when the reservation started (ps).
+        throughput: Application throughput supplied at collection time.
+    """
+
+    owner_name: str
+    storage: InMemoryStorage
+    delivery_owner: str | None = None
+    target_pairs: int = 500
```
hayekr: make this default to `None`

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,281 @@
...
+        """Clear per-owner failure and success counts."""
+        self._failures.clear()
+        self._successes.clear()
+
+
+@dataclass
+class RateMetric(Metric):
+    """Collects a rate value supplied at trial collection time (e.g. throughput)."""
+
+    key: str = "app_throughput"
```
hayekr, line 195: the rate is always `app_throughput`

sequence/utils/metrics/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+        target_pairs: Number of delivered pairs required to compute delivery time.
+        reservation_start_time: Simulation time when the reservation started (ps).
+        throughput: Application throughput to include in collected metrics.
+
+    Returns:
+        Mapping of metric output keys to per-trial values.
+    """
+    ctx = CollectContext(
+        owner_name=owner_name,
+        storage=storage,
```
hayekr, line 135: Please fix the typing on this

sequence/utils/metrics/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+            **record_kwargs,
+        }
+    )
+
+
+def collect_trial_metrics(
+    owner_name: str,
+    *,
+    delivery_owner: str | None = None,
+    target_pairs: int = 500,
```
hayekr: target pairs should not default to 500

sequence/utils/metrics/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+
+    for metric in scalar_metrics:
+        values = [trial[metric] for trial in trials]
+        finite_values = [value for value in values if isinstance(value, (int, float)) and math.isfinite(value)]
+        if finite_values:
+            aggregated[f"avg_{metric}"] = mean(finite_values)
+            aggregated[f"std_{metric}"] = stdev(finite_values) if len(finite_values) > 1 else 0.0
+        else:
+            aggregated[f"avg_{metric}"] = float("nan")
+            aggregated[f"std_{metric}"] = float("nan")
```
hayekr: This is an exact logic copy of lines 181-186

sequence/utils/metrics/builtins.py
```diff
@@ -0,0 +1,61 @@
...
+    success_event=EventTypes.ES_SUCCESS,
+    rate_field="es_success_rate",
+)
+SWAPPED_FIDELITIES_METRIC = FidelityMetric(
+    key="swapped_fidelities",
+    event=EventTypes.ES_SUCCESS,
+    field="fidelity",
+)
+
+_BUILTIN_METRICS = (
```
hayekr: Should the scope of this just be the register_builtin_metrics() method?

pyproject.toml
```diff
@@ -76,3 +76,10 @@ Homepage = "https://github.com/sequence-toolbox/SeQUeNCe"
 Documentation = "https://sequence-rtd-tutorial.readthedocs.io/"
 Issues = "https://github.com/sequence-toolbox/SeQUeNCe/issues"
 Changelog = "https://github.com/sequence-toolbox/SeQUeNCe/blob/master/CHANGELOG.md"
+
```
hayekr, line 79: Remove these, they are covered in the scope fo a new PR

hayekr (changes requested):

pyproject.toml
Tw1ZZLER: Reverted this commit

sequence/utils/metrics/__init__.py
Tw1ZZLER: Now defaults to None

sequence/utils/metrics/metric_types.py
Tw1ZZLER: done

sequence/utils/metrics/storage.py
Tw1ZZLER: Replaced with Timeline

sequence/utils/metrics/__init__.py
Tw1ZZLER: Replaced with Timeline

sequence/utils/metrics/metric_types.py
Tw1ZZLER: Updated the docstrings to be much more descriptive

sequence/utils/metrics/metric_types.py
Tw1ZZLER: Optional methods are now pass

sequence/utils/metrics/__init__.py
Tw1ZZLER: Provided new `_mean_and_std` method to clean up the logic

sequence/utils/metrics/metric_types.py
```diff
@@ -0,0 +1,281 @@
...
+        delivery_records = [
+            record
+            for record in ctx.storage.get_by_owner(delivery_owner)
+            if self.delivery_event is not None and record["event_type"] is self.delivery_event
+        ]
+        delivery_records.sort(key=lambda record: record["sim_time"])
+        if len(delivery_records) < ctx.target_pairs:
+            return {self.key: float("nan")}
+        if ctx.reservation_start_time is None:
+            return {self.key: float("nan")}
```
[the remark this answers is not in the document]
Tw1ZZLER: Already fixed by [695b960](https://github.com/sequence-toolbox/SeQUeNCe/pull/376/commits/695b96045b77c9358099afc769ddb0e307f39592)

sequence/utils/metrics/builtins.py
Tw1ZZLER: Moved BUILTIN_METRICS into `register_builtin_metrics()`

sequence/utils/metrics/__init__.py
Tw1ZZLER: Fixed

sequence/utils/metrics/metric_types.py
Tw1ZZLER: Fixed docstring

sequence/utils/metrics/metric_types.py
Tw1ZZLER: It is always throughput atm, it will get generalized for other Rate-like metrics in the future when they are added. DeliveryTimeMetric currently does something similar but I think in the future if I create other time-based metrics I will generalize that one to include all time / duration metrics.

hayekr (approved):
:shipit:

PR #389, opened 2026-07-02
fix: remove push trigger

hayekr:
ci(dependabot): Update dependabot with correct prefix

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,8 +4,12 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "daily"
+    commit-message:
+      prefix: "chore"
 
   - package-ecosystem: "github-actions"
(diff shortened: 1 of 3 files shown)


caitaozhan (dismissed):

.github/dependabot.yml
caitaozhan, line 1: why is it daily for uv, but weekly for github-actions?

hayekr: github actions updates much slower; they stay stable for much longer

caitaozhan: Just want to make sure we don't get "spammed" by the dependentbots.

hayekr: do you want me to make uv slower? or with a cooldown of 1 month of version release?

caitaozhan (approved):

PR #394, opened 2026-07-03
chore: remove dependency pandas-stubs

hayekr:
Pandas is now type complete, separate stubs package is not needed.


caitaozhan (approved):
Looks good!

PR #392, opened 2026-07-02
ci: Ignore dependabot branches in gitlab mirror

hayekr:
ci: validate_pr title running with wrong permissions on forks

PR #395, opened 2026-07-03
ci: run validate-pr-title on pull_request_target so it works for fork PRs

### LINKED ISSUE #393: validate-pr-title fails for all fork PRs (Resource not accessible by integration)
### Summary
The `validate-pr-title` job in `.github/workflows/pr_checks.yml` fails for pull requests opened from forks:

```
PR Conventional Commit Validation
Error: Resource not accessible by integration
```

This is independent of the PR title being valid — it is a token-permission issue, not a title issue.

### Root cause
The workflow uses `ytanikin/pr-conventional-commits@1.5.2`, whose `add_label` input defaults to `true` (it labels the PR by task type), which requires `pull-requests: write`. The workflow does declare `permissions: pull-requests: write`, but for `pull_request` events triggered from a **fork**, GitHub forces the `GITHUB_TOKEN` to read-only for security. The label write is rejected → `Resource not accessible by integration`, and the job fails.

Net effect: every external (fork) contributor gets a red required check that blocks their PR and makes it look like their title is wrong, even when it is compliant.

### Suggested fix
Set `add_label: 'false'` so the action only validates the title (no write needed):

```yaml
      - name: PR Conventional Commit Validation
        uses: ytanikin/pr-conventional-commits@1.5.2
        with:
          task_types: '["feat","fix","docs","test","ci","refactor","perf","chore","revert"]'
          add_label: 'false'
```

Alternatively, keep labels for same-repo PRs and skip the label step on forks via `if: github.event.pull_request.head.repo.fork == false`.

Happy to open a PR with the one-line change if welcome. Context: ran into this on #391.

rayair250-droid:
Fixes #393.

Fork PRs receive a read-only \`GITHUB_TOKEN\`, so the label-writing step of \`ytanikin/pr-conventional-commits\` fails with \`Resource not accessible by integration\` for every external contributor — the required check goes red regardless of the (valid) title.

Following @hayekr's suggestion in #393, this switches the trigger from \`pull_request\` to \`pull_request_target\`, so the workflow runs with the base repository's token and can apply the label. The job only reads the PR title from the event payload and does not check out or execute PR code, so `pull_request_target` is safe here.

Note: this PR is itself a fork PR, so its own \`validate-pr-title\` run will still hit the old behavior until this is merged.

### DIFF
--- a/.github/workflows/pr_checks.yml
+++ b/.github/workflows/pr_checks.yml
@@ -1,6 +1,6 @@
 name: Validate PRs
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize, reopened, edited]
 
 permissions:


hayekr:
This change is already in https://github.com/sequence-toolbox/SeQUeNCe/pull/392

Thank you though!

PR #404, opened 2026-07-07
feat: Metrics tracking for network management

Tw1ZZLER:
Add new EventTypes `RESERVATION_APPROVED`, `RESERVATION_REJECTED`, and `RESERVATION_COMPLETION`, as well as `THROUGHPUT` for the modified throughput metric.  Throughput and reservation start time have been removed from the minimized `CollectContext` (that we are trying to do away with altogether) and are now based on existing metrics. It is worth noting that the collect method of `RateMetric` has been adjusted to quite literally copy **and completely replace** `RequestApp.get_throughput()`. It has also been renamed to `ThroughputMetric` as result.


Tw1ZZLER:
Ready for review just gonna add a few more `metrics.record()` call sites but that its it

sequence/utils/metrics/builtins.py
```diff
@@ -24,26 +19,41 @@
...
+PURIFIED_FIDELITIES_METRIC = EventAttributeMetric(
+    key="purified_fidelities",
+    event=EventTypes.EP_SUCCESS,
+    extractor=lambda record: record.data.fidelity,
+)
+SWAPPED_FIDELITIES_METRIC = EventAttributeMetric(
     key="swapped_fidelities",
     event=EventTypes.ES_SUCCESS,
-    field="fidelity",
+    extractor=lambda record: record.data.fidelity,
```
hayekr, line 36: I find the use of lambda here odd, but I think its okay

sequence/utils/metrics/event_types.py
```diff
@@ -13,42 +13,33 @@ class EventType:
 
     Attributes:
         name: Unique string identifier for the event type.
+        payload_type: Dataclass type for the event payload, or `None`
+            if the event carries no additional data beyond event type
+            and owner name.
     """
 
     name: str
+    payload_type: type | None = None
```
hayekr, line 23: I would add your guarding that you do in record() as a __post_init__ here.

sequence/network_management/routing/routing_distributed.py
```diff
@@ -468,7 +470,11 @@ def set_state(self, neighbor: str, new_state: str):
                 fsm.reset()
                 if neighbor in self.adj_cost:
                     del self.adj_cost[neighbor]
+                    metrics.record(EventTypes.NEIGHBOR_DOWN, self.owner.name, neighbor=neighbor)
                     self.originate_and_flood()
+            case "Full":
```
hayekr: Unsure if there was a reason to not include this state of OSPF originally. @caitaozhan ?

hayekr (dismissed):
Looks good. Couple comments.

sequence/utils/metrics/builtins.py
Tw1ZZLER: I refactored the metric type here to be much more general, so instead of just passing in the field, you specify how that field is obtained from all attributes of the specified record. With this method, you can also get the other record attributes, namely simulation time, owner name, and the event type (although the event type would be very useless in this context)

It takes any callable function, but my vision is that users just use a simple lambda to extract the correct field they want from the record object.

sequence/network_management/routing/routing_distributed.py
Tw1ZZLER: If not, where would you suggest the NEIGHBOR_FULL event occurs?

sequence/network_management/routing/routing_distributed.py
```diff
@@ -410,6 +411,7 @@ def expire_lsa(self, last_originated_time: int):
         updated = self.lsdb.install(withdrawal, now)
         if updated:
             forwarding_table = self.run_spf()
+            metrics.record(EventTypes.ROUTE_RECOMPUTED, self.owner.name, num_routes=len(forwarding_table))
```
caitaozhan: What metric is this? Time, throughput related?

Tw1ZZLER: Removed

sequence/network_management/routing/routing_distributed.py
Tw1ZZLER: Removed

hayekr (approved):
lgtm

PR #418, opened 2026-07-13
feat: Add App abstraction class to standardize API and add hooks

Tw1ZZLER:
This PR adds a new abstract class `App` that other apps such as `RequestApp` and `PeriodicApp` now inherit from. It includes 3 new abstract methods that are required to be meaningfully implemented in the subclasses of App. These are `get_memory()`, `get_reservation_result()`, and `get_other_reservation()`.

### DIFF
--- a/sequence/app/__init__.py
+++ b/sequence/app/__init__.py
@@ -1,4 +1,5 @@
-__all__ = ['app', 'random_request']
+__all__ = ["app", "random_request", "teleport_app", "request_app"]
+
 
 def __dir__():
     return sorted(__all__)
--- a/sequence/app/__init__.py
+++ b/sequence/app/__init__.py
@@ -1,4 +1,4 @@
-__all__ = ['random_request']
+__all__ = ['app', 'random_request']
 
 def __dir__():
     return sorted(__all__)
--- a/sequence/app/app.py
+++ b/sequence/app/app.py
@@ -35,13 +35,13 @@ class App(ABC):
         name (str): Human-readable name for the application instance.
     """
 
-    def __init__(self, node: "QuantumRouter"):
-        self.node: "QuantumRouter" = node
+    def __init__(self, node: QuantumRouter):
+        self.node: QuantumRouter = node
         self.node.set_app(self)
         self.name: str = f"{self.node.name}.{self.__class__.__name__}"
 
     @abstractmethod
-    def get_memory(self, info: "MemoryInfo") -> None:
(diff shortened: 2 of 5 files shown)


sequence/app/app.py
```diff
@@ -0,0 +1,92 @@
...
+    The constructor automatically registers the application on the node via
+    `node.set_app(self)`, so subclasses should always call
+    `super().__init__(node)`!
+
+    Attributes:
+        node (QuantumRouter): The node this application is attached to.
+        name (str): Human-readable name for the application instance.
+    """
+
+    def __init__(self, node: "QuantumRouter"):
```
hayekr: When you use annotations you dont need the quotes

sequence/app/app.py
```diff
@@ -0,0 +1,92 @@
...
+        self.node.set_app(self)
+        self.name: str = f"{self.node.name}.{self.__class__.__name__}"
+
+    @abstractmethod
+    def get_memory(self, info: "MemoryInfo") -> None:
+        """Called when an entangled memory becomes available.
+
+        Args:
+            info (MemoryInfo): Information about the available memory.
+        """
```
hayekr, line 49: I would make sure this either uses the `pass` instruction or `...`

sequence/app/app.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @abstractmethod
+    def get_reservation_result(self, reservation: "Reservation", result: bool) -> None:
+        """Called when a reservation result is received from the network manager.
+
+        Args:
+            reservation (Reservation): The reservation that completed.
+            result (bool): Whether the reservation was approved.
+        """
```
hayekr, line 59: Use `pass` or `...`. Same for every other abstract method

Tw1ZZLER:
Removed `get_throughput()`. If the user wants throughput from the app they should use `THROUGHPUT_METRIC`. Ideally `get_throughput()` should be deprecated

sequence/app/__init__.py
caitaozhan, line 1: Let's add `request_app` and `teleport_app`.

Tw1ZZLER: done

Tw1ZZLER:
Added warning that `get_throughput()` is deprecated and `THROUGHPUT_METRIC` should be used instead. 

should be ready for merge :slightly_smiling_face:

hayekr (approved):

PR #419, opened 2026-07-14
feat: change metrics.enable to use metrics instead of event types

Tw1ZZLER:
Changes the `metrics.enable()` method to take a list of `Metric` instead of a list of `EventType`, which makes a lot more sense for the end user. Edits the `_enabled_events` set and adds the EventTypes that each enabled metric requires to be recorded.

Also exports all the builtin metrics globally which makes it easier for the user to use them


hayekr (approved):
:shipit:

PR #426, opened 2026-07-17
feat: support for reservation early expire

caitaozhan:
1. Add new function `NetworkManager.remove_reservation_from_timecards()`. When a reservation expires at `T` before its reservation `end_time`, remove the reservation from the timecards, so that the network manager can schedule new incoming reservations between `[T, end_time]` for that memory.
2. In `ResourceManager.generate_load_rules()`, remove the `update` events that set memory to `RAW`
3. In `RuleManager.expire()`, return `None` when the rule does not exist (e.g., in the case of early expiration)
4. In `ResourceManager.expire()`, set the memory to `RAW` (do the job in 2.), and adjust to case 3.
5. Update priority in `ResourceManager.generate_load_rules()`. Previously, EG, EP, and ES rules all had the same priority 10. Now EG's priority is 10 (higher priority), EP is 20, and ES's is 20 (lower)
6. Add `test_EarlyExpire()` and update `test_expire()` in pytest


sequence/resource_management/resource_manager.py
```diff
@@ -261,12 +270,13 @@ def generate_load_rules(self, path: list[str], reservation: Reservation, timecar
...
-                                  [None, self.owner.components[memory_array_name][card.memory_index], "RAW"])
-                event = Event(reservation.end_time, process, self.owner.timeline.schedule_counter)
-                self.owner.timeline.schedule(event)
+        # NOTE: the following is covered by the "expire" process above. Thus, not needed.
+        # for card in timecards:
+        #     if reservation in card.reservations:
+        #         process = Process(self.owner.resource_manager, "update",
+        #                           [None, self.owner.components[memory_array_name][card.memory_index], "RAW"])
+        #         event = Event(reservation.end_time, process, self.owner.timeline.schedule_counter)
+        #         self.owner.timeline.schedule(event)
```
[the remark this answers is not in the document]
caitaozhan: Let the commented code exist for a while

sequence/network_management/network_manager.py
```diff
@@ -131,6 +131,18 @@ def generate_rules(self, reservation: Reservation):
...
+        """Remove a reservation from the timecards. 
+        
+        Useful when a request is served before the end_time and then early expired the reservation.
+
+        Args:
+            reservation (Reservation): reservation to remove from timecards.
+        """
+        for timecard in self.timecards:
+            if reservation in timecard.reservations:
+                timecard.remove(reservation)
```
hayekr: This function can be O(n) but is O(n^2). timecard.remove() already guards against a missing reservation. The conditional here isnt needed, you can just loop through timecards and call timecard.remove(reservation)

sequence/resource_management/resource_manager.py
```diff
@@ -171,12 +171,13 @@ def generate_load_rules(self, path: list[str], reservation: Reservation, timecar
         index: int = path.index(self.owner.name)
 
         # Create Rules
+        priority = 10
```
hayekr: We should just have three constants. EG_PRIORITY, ES_PRIORITY, EP_PRIORITY instead of having a mutating variable representing all three.

sequence/resource_management/resource_manager.py
```diff
@@ -261,12 +264,13 @@ def generate_load_rules(self, path: list[str], reservation: Reservation, timecar
             event = Event(reservation.end_time, process, self.owner.timeline.schedule_counter)
             self.owner.timeline.schedule(event)
 
-        for card in timecards:
-            if reservation in card.reservations:
-                process = Process(self.owner.resource_manager, "update",
-                                  [None, self.owner.components[memory_array_name][card.memory_index], "RAW"])
-                event = Event(reservation.end_time, process, self.owner.timeline.schedule_counter)
-                self.owner.timeline.schedule(event)
+        # NOTE: the following is covered by the "expire" process above. Thus, not needed.
```
hayekr: If not needed, delete. We should not keep commented out code in prod.

sequence/resource_management/resource_manager.py
```diff
@@ -319,6 +323,15 @@ def expire(self, rule: Rule) -> None:
             for memory in protocol.memories:
                 self.update(protocol, memory, MemoryInfo.RAW)
 
+        # below do the job of line 264~270 (commented code) in generate_load_rules, which updates the memory to RAW when the reservation is expired.
```
hayekr: Make this comment clear. We shouldn't refer back to commented code. Commented out code should be removed

hayekr (changes requested):
This PR completes the support for early expiration of reservations. Minor changes include some fixes on logging and correcting import method. This is a needed change, needs some minor polish. Requesting Changes

sequence/network_management/network_manager.py
caitaozhan: `timecard.remove(reservation)` still costs `O(n)`, making the function `O(n^2)`.

sequence/resource_management/resource_manager.py
caitaozhan: will delete it.

hayekr (approved):
This PR completes the support for early expiration of reservations. :shipit:

PR #429, opened 2026-07-21
feat: Change ruff line length for formatting to 120 chars

Tw1ZZLER:
Changes line length for the ruff formatter to 120 chars in the `pyproject.toml` file


Tw1ZZLER:
will be implemented later

PR #438, opened 2026-08-05
feat: Metrics tracking for resource management + application, metrics from literature

Tw1ZZLER:
This PR is rather large and monolithic, but its to avoid Git madness.
This PR depends on #404 

Adds new events `MEMORY_EXPIRE`, `MEMORY_UPDATE`, `SWAP_FIDELITY_VIOLATION`,`DELIVERY_FIDELITY_VIOLATION`, and finally implements `DELIVERY`. Implementation of `DELIVERY` also allows for the removal of `CollectContext`, and a helper can be created to get the information from storage without the user having to pass it when calling `collect_trial_metrics`.

Implements the following new metrics from literature:
- [Admission rate](https://ieeexplore.ieee.org/document/9964002)
- [Bell pair utilization ratio](https://arxiv.org/html/2505.12459v2#:~:text=III%2DDBell%20pair%20utilization%20rate)
- [**Time to Serve**, or On-demand end-to-end latency](https://arxiv.org/html/2407.11376v2#:~:text=Definition%2010%20%28Latency%29%2E)
- [Memory utilization ratio](https://arxiv.org/html/2603.27570v1#:~:text=The%20memory%20utilization%20ratio%20is%20then%20Qv%2FMv%E2%88%88%5B0%2C1%5D%2E%20Under%20a%20best%2Deffort%20delivery%20model%2C%20we%20assign%20uniform%20weights%20to%20all%20requests%20%E2%9F%A8s%2Cd%E2%9F%A9%E2%88%88R%2E)
- [Memory decoherence rate](https://arxiv.org/html/2603.27570v1)
- [Reservation success rate](https://arxiv.org/html/2505.12459v2#:~:text=III%2DCSuccess%20rate)
- [Fidelity threshold violation rate](https://arxiv.org/html/2501.03210v1#:~:text=the%20fidelity%20threshold) ([also in this paper](https://arxiv.org/pdf/2204.09538v2))
    - Implemented for deliveries
- [Multi-tenant fairness, also called Jain's Fairness Index](https://arxiv.org/html/2603.27570v1#bib.bib7:~:text=using%20Jain%E2%80%99s%20Fairness%20Index)


sequence/app/request_app.py
```diff
@@ -135,9 +136,40 @@ def get_memory(self, info: MemoryInfo) -> None:
 
         if info.index in self.memo_to_reservation:
             reservation = self.memo_to_reservation[info.index]
-            if info.remote_node == reservation.initiator and info.fidelity >= reservation.fidelity:
+            remote_is_peer = (
+                info.remote_node == reservation.initiator
+                or info.remote_node == reservation.responder
+            )
```
hayekr, line 142: This will always be True. A remote node is always the initiator or responder. What case are you guarding for?

sequence/app/request_app.py
```diff
@@ -135,9 +136,40 @@ def get_memory(self, info: MemoryInfo) -> None:
 
         if info.index in self.memo_to_reservation:
             reservation = self.memo_to_reservation[info.index]
-            if info.remote_node == reservation.initiator and info.fidelity >= reservation.fidelity:
+            remote_is_peer = (
```
hayekr, line 139: Revert the conditional to the way it was. Add an else that records a violation. This will catch all cases. The other branches will only be false on insufficient fidelity.

sequence/app/teleport_app.py
```diff
@@ -76,6 +77,30 @@ def get_memory(self, info: MemoryInfo):
         # once we see our entangled half, hand it to the protocol
         if info.index in self.memo_to_reservation:
             if info.state == "ENTANGLED":
+                reservation = self.memo_to_reservation[info.index]
+                if info.fidelity < reservation.fidelity:
```
hayekr, line 81: This protocol doesn't compare the fidelity. Also get_memory here does not consume the pair; notice there is no setting to 'RAW'

hayekr:
@caitaozhan General question abou tthis. Do we want to record these violations?

sequence/entanglement_management/swapping/swapping_bds.py
```diff
@@ -96,6 +96,18 @@ def start(self) -> None:
                 right_node=self.right_node,
                 fidelity=fidelity,
             )
+            if (
```
hayekr, line 99: @caitaozhan See violation record example

hayekr (dismissed):
See comments. Thanks!

sequence/entanglement_management/swapping/swapping_bds.py
caitaozhan: Why do we want to record this?

caitaozhan: Also, the `if` statement is very ugly -- it should not occupy 5 lines of code.

hayekr:
Current plan is to remove the added metrics and most recording. Pear down the PR to just include resource/application metrics tracking features. Metrics will be evaluated at a later date.

hayekr:
Further, I am moving the metrics source out of utils

hayekr:
Reopening

sequence/network_management/reservation.py
```diff
@@ -6,8 +6,10 @@
 """
 from __future__ import annotations
 
+from typing import Mapping, Any
 
-class Reservation:
+
+class Reservation(Mapping[str, Any]):
```
caitaozhan, line 12: This is a good idea because it allows using `**` in kwargs.

In `DistributedNetworkManager.pop()`, the `metrics.record()` should also be simplified using `**`

sequence/network_management/reservation.py
```diff
@@ -22,6 +24,18 @@ class Reservation:
...
 
+    _EVENT_FIELDS = {
+        "identity": "identity",
+        "initiator": "initiator",
+        "responder": "responder",
+        "start_time": "start_time",
+        "end_time": "end_time",
+        "memory_size": "memory_size",
+        "entanglement_number": "entanglement_number",
+        "target_fidelity": "fidelity",
```
caitaozhan, line 35: Shall we do ` "target_fidelity": "target_fidelity"`? Debatable, but if so, we need to rename the attribute.

caitaozhan (commented):
Looks good. Left two comments.

hayekr (approved):

PR #440, opened 2026-08-11
fix: guard draw_topo CLI behind a main() entry point

### LINKED ISSUE #436: sequence/utils/draw_topo.py runs parse_args() at import time, breaking its autodoc page (draw_topo.rst) on the docs site
sequence/utils/draw_topo.py executes parser.parse_args() (and the entire CLI body, file I/O and graph drawing) at module import time, instead of under an if __name__ == "__main__": guard.
Impact
- Importing sequence.utils.draw_topo fails with SystemExit (argparse prints usage and exits when no config_file is supplied).
- Because of this, Sphinx autodoc cannot document the module: the Module References → Miscellaneous → Utils page docs/source/references/misc/utils/draw_topo.rst renders empty on the deployed Read the Docs site.
- The module can't be imported programmatically either (e.g., for testing or reuse).

Proposed fix
Guard the CLI entry point in sequence/utils/draw_topo.py:
```python
def main() -> None:
    parser = argparse.ArgumentParser()
    ...
    args = parser.parse_args()
    ...  # existing body

if __name__ == "__main__":
    main()
```
This is a low-risk, self-contained change. I can open a PR if helpful.

eastagiletracker:
This PR proposes guarding the `sequence/utils/draw_topo.py` command line interface behind a `main()` entry point, so that importing the module no longer exits the interpreter (Fixes #436). We include this PR work along with a full history of your repo at https://eastagiletracker.com/projects/207. You can sign in with your GitHub ID to claim ownership of the project.

## What is wrong

`sequence/utils/draw_topo.py` calls `parser.parse_args()` at module scope, together with the rest of the command line body: reading the config file, constructing the topology and building the graph all happen as a side effect of importing the module. Because `argparse` exits when the required `config_file` argument is absent, importing the module raises `SystemExit` before a caller can do anything with it. On `master` at `09cfb67`:

```console
$ python -c "import sequence.utils.draw_topo"
usage: -c [-h] [-d DIRECTORY] [-f FILENAME] [-m] config_file
-c: error: the following arguments are required: config_file
$ echo $?
2
```

The consequence reaches a little further than the reference page described in #436. Since the module terminates the interpreter that imports it, `pytest` cannot collect a test module that imports it either — the whole run aborts during collection with `INTERNALERROR ... SystemExit: 2` instead of reporting a failure, which is why this module has no tests today. `docs/scripts/generate_reference_rst.py` generates a `misc/utils` reference section from every module in `sequence/utils`, so autodoc imports this one as well.

## The change

The module body is split into three functions with no behaviour change: `get_topology()` resolves a config file to the matching `RouterNetTopo`, `QKDTopo` or `DQCNetTopo`, `draw_topology()` builds the graphviz graph, and `main()` does the argument parsing and rendering. `main()` runs only under `if __name__ == "__main__"`, so `python -m sequence.utils.draw_topo` and `python sequence/utils/draw_topo.py` keep working exactly as before, with the same flags and the same output, while `import sequence.utils.draw_topo` is now free of side effects. Splitting the drawing out of the command line body is also what makes the module testable, and it lets the graph be built programmatically, which #436 asks for.

## How this was verified

The full suite was run on a clean checkout before the change and after it: 407 tests pass on unmodified `master` and 414 pass with this change applied, with nothing moving from passing to failing. The seven new tests in `tests/utils/test_draw_topo.py` cover the side-effect-free import, the topology dispatch for each supported config type, the unknown node type path, and both the default and `--draw_middle` drawing branches; against the unmodified module they cannot even be collected, failing with the `SystemExit` above. The command line itself was exercised end to end against `tests/topology/router_net_topo_sample_config.json`, and the graph it emits is unchanged:

```
graph {
	layout=neato overlap=false
	router_0
	router_1
	router_2
	router_0 -- router_1 [color=blue]
	router_1 -- router_2 [color=blue]
}
```

`ruff check` reports the same findings on `sequence/utils/draw_topo.py` as it does on `master`, and none on the new test module.

## How this was managed

We imported your issues and pull requests into a live agile board and used it to track this work: the story for #436 is at https://eastagiletracker.com/projects/207/stories/75053, on the board at https://eastagiletracker.com/projects/207 (425 stories and 18 labels imported from this repository).

board

If you'd rather not receive contributions like this, reply `no-more-prs` on this pull request and we won't open any further ones on your repositories.

---

**Lawrence W. Sinclair**
CEO / East Agile
[linkedin.com/in/lwsinclair/](https://linkedin.com/in/lwsinclair/)
[eastagile.com](https://eastagile.com)

### DIFF
--- a/sequence/utils/draw_topo.py
+++ b/sequence/utils/draw_topo.py
@@ -11,71 +11,94 @@ from graphviz import Graph
 from json import load
 from collections import defaultdict
 
+from sequence.topology.topology import Topology
 from sequence.topology.router_net_topo import RouterNetTopo
 from sequence.topology.qkd_topo import QKDTopo
 from sequence.topology.dqc_net_topo import DQCNetTopo
 
 
-parser = argparse.ArgumentParser()
-parser.add_argument('config_file', help="path to json file defining network")
-parser.add_argument('-d', '--directory', type=str, default='tmp', help='directory to save the figure')
-parser.add_argument('-f', '--filename', type=str, default='topology', help='filename of the figure')
-parser.add_argument('-m', '--draw_middle', action='store_true')
+def get_topology(config_file: str) -> Topology:
+    """
+    Determine the type of network described by a config file and load it.
+    Args:
+        config_file: path to json file defining network
 
-args = parser.parse_args()
-directory = args.directory
-filename = args.filename
-draw_middle = args.draw_middle
+    Returns: The topology matching the type of the config file's first node
+    """
+    with open(config_file, 'r') as fh:
+        config = load(fh)
+    nodes = config["nodes"]
+    node_type = nodes[0]["type"]
 
-# Determine type of network
-with open(args.config_file, 'r') as fh:
-    config = load(fh)
-nodes = config["nodes"]
-node_type = nodes[0]["type"]
+    if node_type == RouterNetTopo.BSM_NODE or node_type == RouterNetTopo.QUANTUM_ROUTER:
+        return RouterNetTopo(config_file)
 
-if node_type == RouterNetTopo.BSM_NODE or node_type == RouterNetTopo.QUANTUM_ROUTER:
-    topo = RouterNetTopo(args.config_file)
+    elif node_type == QKDTopo.QKD_NODE:
+        return QKDTopo(config_file)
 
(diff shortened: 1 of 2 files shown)


hayekr:
no-more-prs

PR #443, opened 2026-08-13
refactor: faster quantum ket manager and remove QuantumManagerDenseQubit

ssuresh27-byte:
- A faster implementation of `QuantumManagerKet` -- from O(4^k) down to O(2^k), where k is the number of qubits
- Got rid of `QuantumManagerDenseQubit` as most of the functionalities are not being used in ket_vector (such as prepare_circuit). 
  - Future plan: get rid of prepare_circuit entirely as density_matrix can be sped up using the same method as the ket_vector.
- update the related pytest, including fixing the `correct_order()` in `test_purification.py` and reducing the runtime of the `test_cascade()`
- Add a new module `kernel.quantum_gates` to store all the matrices of the quantum gates.
- Add a new module `kernel.quantum_manager.utils` to store some common functions within the quantum manager.
- Removed some legacy code


caitaozhan (dismissed):
looks good to me

Alex-Kolar (dismissed):
I agree that the merge looks good, but I have a few comments for future development:

1. What is the transient BSM behavior mentioned in the `test_quantum_manager` file? I'm not sure if I follow the description
2. Some of the circuit matrix definitions seems a bit redundant between `circuit.py` and the new `ket_vector.py` quantum manager. We might want to consider consolidating these in the future.
3. I would say there is a similar problem between the Pauli error specification for the bell diagonal state formalism and the newly implemented errors for the density matrix formalism, although I'm not sure how easy these would be to consolidate.

sequence/kernel/quantum_manager/base.py
```diff
@@ -152,101 +151,41 @@ def set_states(self, states: dict):
...
-    - Stabilizer: qubit + tableau, not dense vector/matrix.
-    - Bell diagonal: qubit-pair state with compact Bell-basis probabilities, not dense vector/matrix.
-    - Fock density: dense matrix, but not restricted to qubit subsystems.
+    Args:
+        circuit (Circuit): quantum circuit to apply.
+        keys (list[int]): list of keys for quantum states to apply circuit to.
+        meas_samp (float): random sample used for measurement.
     """
+    if len(keys) != circuit.size:
+        raise ValueError("mismatch between circuit size and supplied qubits")
```
hayekr: Nit-picks but can we please capitalize and have proper punctuation in error messages.

sequence/kernel/quantum_manager/base.py
```diff
@@ -152,101 +151,41 @@ def set_states(self, states: dict):
         self.states = states
```
hayekr: These module level functions that are not part of the base class violate single responsibility principle. These helpers should not be in `base.py` and should be in a new `utils.py` for proper sep of concerns.

sequence/kernel/quantum_manager/density_matrix.py
```diff
@@ -3,25 +3,98 @@
...
+
+        The infidelity -> channel-strength conversion matches the stabilizer manager:
+        for an average gate fidelity F on a d-dimensional system the total error
+        probability is (d+1)/d * (1 - F): 1.5*(1-F) for 1-qubit gates (d=2) and
+        1.25*(1-F) for 2-qubit gates (d=4), capped at 1.
+        """
         super().__init__()
+        self.noise_rng = np.random.default_rng(seed)
+        self.noise_type: str = kwargs.get("noise_type", "depolarize")
+        if self.noise_type not in self._NOISE_ALPHABET:
```
hayekr: I think `_noise_enabled()` be exec in the constructor to validate the passed noise type instead of in `run_circuit()`? The output of this should also just be set as a class variable, i.e., (`self.noise_enabled = _noise_enabled()`)

sequence/kernel/quantum_manager/base.py
```diff
@@ -152,101 +151,41 @@ def set_states(self, states: dict):
...
-                gate = Gate("SWAP", targets=[i, j])
-                swap_circuit.add_gate(gate)
-                all_keys[i], all_keys[j] = all_keys[j], all_keys[i]
-        swap_mat = gate_sequence_product(swap_circuit.propagators()).full()
-        return all_keys, swap_mat
+    Returns:
+        tuple: updated list of all keys and the swap matrix.
+    """
+    swap_circuit = QubitCircuit(N=len(all_keys))
+    for i, key in enumerate(keys):
```
hayekr: What is a typical size of `keys` and `all_keys`? If these ca be quite large we can speed this function up.

sequence/kernel/quantum_manager/density_matrix.py
```diff
@@ -63,6 +142,328 @@ def run_circuit(self, circuit: Circuit, keys: list[int], meas_samp=None) -> dict
...
+        old_states: list[np.ndarray] = []
+        all_keys: list[int] = []
+        for key in keys:
+            qstate = self.states[key]
+            if qstate.keys[0] not in all_keys:
+                old_states.append(np.asarray(qstate.state, dtype=complex))
+                all_keys += list(qstate.keys)
+        rho = np.array([[1.0 + 0j]])
+        for state in old_states:
+            rho = np.kron(rho, state)
```
hayekr: this is the same as quantum_utils.kron? Should we just stop using quantum_utils to pass numpy functions?

sequence/kernel/quantum_manager/density_matrix.py
```diff
@@ -63,6 +142,328 @@ def run_circuit(self, circuit: Circuit, keys: list[int], meas_samp=None) -> dict
...
+        for key in keys:
+            qstate = self.states[key]
+            if qstate.keys[0] not in all_keys:
+                old_states.append(qstate.state)
+                all_keys += qstate.keys
+
+        # construct compound state; order qubits
+        new_state = [1]
+        for state in old_states:
+            new_state = kron(new_state, state)
```
hayekr: There is no append here, so new_state will always have one element equal to `kron(new_state, state[-1])`

hayekr: Also I think a move to np.kron is better.

sequence/qkd/cascade.py
```diff
@@ -416,8 +416,7 @@ def generate_key(self, keylen: int, frame_num=math.inf, run_time=math.inf) -> No
 
         log.logger.info(self.name + f' state={self.state} generate_key, keylen={keylen}, keynum={frame_num}')
         if self.role == 1:
-            raise Exception(
-                f"Cascase.generate_key() called on receiver '{self.name}'")
+            raise Exception(f"Cascase.generate_key() called on receiver '{self.name}'")
```
hayekr: Cascade is spelled wrong...

hayekr (changes requested):
Only minor suggestions.

sequence/kernel/quantum_manager/density_matrix.py
caitaozhan: Yes. Fixed it.

sequence/kernel/quantum_manager/density_matrix.py
caitaozhan: Fixed. Now only np.kron

sequence/kernel/quantum_manager/base.py
caitaozhan: len(keys) is always 2.
len(all_keys) is usually less than or equal to 4 in network protocols. But it can exceed 10 when executing distributed quantum circuits.

caitaozhan:
> I agree that the merge looks good, but I have a few comments for future development:
> 
> 1. What is the transient BSM behavior mentioned in the `test_quantum_manager` file? I'm not sure if I follow the description
> 2. Some of the circuit matrix definitions seems a bit redundant between `circuit.py` and the new `ket_vector.py` quantum manager. We might want to consider consolidating these in the future.
> 3. I would say there is a similar problem between the Pauli error specification for the bell diagonal state formalism and the newly implemented errors for the density matrix formalism, although I'm not sure how easy these would be to consolidate.

About the redundancy between `circuit.py` and the new `ket_vector.py`
- The fast operations in `ket_vector` rely on all sorts of numpy operations and circumvent the `*_gate()` functions that return a QuTiP `Qobj` object.
- Redundancy may be reduced if the matrices are defined globally.

sequence/kernel/quantum_manager/density_matrix.py
caitaozhan: fixed.

sequence/kernel/quantum_manager/base.py
caitaozhan: fixed in https://github.com/sequence-toolbox/SeQUeNCe/pull/443/changes/166e9bb2080b5ab2f5bc4cb93a2116fa008b155a

caitaozhan:
> I agree that the merge looks good, but I have a few comments for future development:
> 
> 1. What is the transient BSM behavior mentioned in the `test_quantum_manager` file? I'm not sure if I follow the description
> 2. Some of the circuit matrix definitions seems a bit redundant between `circuit.py` and the new `ket_vector.py` quantum manager. We might want to consider consolidating these in the future.
> 3. I would say there is a similar problem between the Pauli error specification for the bell diagonal state formalism and the newly implemented errors for the density matrix formalism, although I'm not sure how easy these would be to consolidate.

The redundancy issue is resolved in https://github.com/sequence-toolbox/SeQUeNCe/pull/443/changes/7a6f0085e82dc4d099138495e8200a8ebebfc1f2.

For that transient BSM in `test_quantum_manager` comment, I need to check with Sanjay.

caitaozhan:
> I agree that the merge looks good, but I have a few comments for future development:
> 
> 1. What is the transient BSM behavior mentioned in the `test_quantum_manager` file? I'm not sure if I follow the description
> 2. Some of the circuit matrix definitions seems a bit redundant between `circuit.py` and the new `ket_vector.py` quantum manager. We might want to consider consolidating these in the future.
> 3. I would say there is a similar problem between the Pauli error specification for the bell diagonal state formalism and the newly implemented errors for the density matrix formalism, although I'm not sure how easy these would be to consolidate.

That transient BSM behavior pytest is removed.

sequence/kernel/quantum_manager/density_matrix.py
caitaozhan: The noise-related code in this PR is removed

hayekr (approved):
:ship:

PR #455, opened 2026-09-04
ci: finalize generated changelog format

hayekr:
This PR finalizes the changelog format that PSR generates. It also ignores the PR's tagged `chore:` as they are mainly used by dependabot and are "noisy"


caitaozhan (approved):
LGTM

PR #458, opened 2026-09-12
refactor: Remove GUI

caitaozhan:
The current GUI is removed because:
- The only working function is generating config files, and this functionality can be done better with a new `CLI` tool.
- The `run` function is a work in progress (see the `WIP` in the screenshot below) and runs into bugs.
- Nobody wants to fix it for years. Not maintainable.
- Thus, let's remove the current `gui` and redo it from scratch as an independent repository (a good SULI project).
- After removing the `gui`, the number of dependent packages is reduced by 10+.

image

---

The removed code is transferred to [SeQUeNCe-Examples](https://github.com/sequence-toolbox/SeQUeNCe-Examples)

This PR also removes the CI badge.


hayekr (approved):
lgtm

